PR #1, opened 2016-12-07
Ecchen/cmakesystem

grimmerk:
Modification: 

1.  create simple cmake 
2.  fix .pro
3. fix include path


carta/cpp/plugins/plugins.pro
```diff
@@ -4,35 +4,35 @@ TEMPLATE = subdirs
 SUBDIRS += casaCore-2.10.2016
 #SUBDIRS += casaCore-2.0.1
 SUBDIRS += CasaImageLoader
-SUBDIRS += Colormaps1
```
grimmerk, line 7: Before merging to develop, these lines which disable plugins need to recover enable back. Right?

grimmerk:
The conflict should be solved first (the main conflict is CircleCI.yml needed to be added back by using develop version), otherwise the merge of this pull request can not be done successfully. After thinking, it may be better to divide this branch to the following parts. 
A. a small set of fix paths which is needed for cmake and qmake. 
B. the currnet non-finshed Cmake modification (B can include A). 
C. new fixes for paths in the future.  We can merge A part to develop. Then remain B part. If any new fixes for paths, create a small branch for them. 

Another question is, the fix for the same (plugin) path for qmake and cmake are two places/commits, or the same line/commit?

carta/CMakeLists.txt
```diff
@@ -0,0 +1,118 @@
...
+set(CARTAPROJECT_PATCH_VERSION 0)
+set(CARTAPROJECT_VERSION ${CARTAPROJECT_MAJOR_VERSION}.${CARTAPROJECT_MINOR_VERSION}.${CARTAPROJECT_PATCH_VERSION})
+message(STATUS "CartaProject version " ${CARTAPROJECT_VERSION})
+
+# find package
+find_package(Qt5Core REQUIRED)
+#find_package(Qt5Gui REQUIRED)
+#find_package(Qt5Network REQUIRED)
+#find_package(Qt5Widgets REQUIRED)
+#find_package(Qt5WebKitWidgets REQUIRED)
```
grimmer0125: We use qtwebkit in carta, so is it safe not to #find_package(Qt5WebKitWidgets REQUIRED)?

grimmer0125:
test2

grimmerk:
@CNOCycle , since this is in pending status, so let us close this, if there is any further update in the future, create a new one.

keflavich:
Right, yes, the difference between "folders that are files" and "folders that should be browsed to" makes the problem trickier than I implied!  Still, just showing them all, then deciding later if they're images or not, would be nice.

PR #6, opened 2017-01-03
re-use gridline

grimmerk:
don't calculate grid again when switching between different  channels.


grimmerk:
@kswang1029 , @CNOCycle  From Susan:   I guess with the re-use gridline issue (6), I would worry that the display axes would get permuted and the grid would not get redrawn, but I assume you have checked and that does not happen.

PR #9, opened 2017-01-12
Ecchen/improved rendering

CNOCycle:
1. new cache in rendering function
- Redesign the cache mechanism and cache the raw GRB image.

2. PixelPipeline performance improved
- performance improved by auto-rounding

### DIFF
--- a/carta/cpp/CartaLib/PixelPipeline/IPixelPipeline.h
+++ b/carta/cpp/CartaLib/PixelPipeline/IPixelPipeline.h
@@ -34,7 +34,7 @@ namespace PixelPipeline
 ///   - can have mulitple sub-steps
 ///   - configurable
 /// - stage 5: Drgb -> QRgb
-///   - simple conversion to 8 bit values using rounding
+///   - simple conversion to 8 bit values using auto-rounding
 ///   - not configurable
 ///
 /// Setup for a changed min/max
@@ -258,7 +258,7 @@ public:
         convert( p_val, drgb );
 
         /// stage 5:
-        result = qRgb( round( drgb[0] * 255 ), round( drgb[1] * 255 ), round( drgb[2] * 255 ) );
+        result = qRgb( drgb[0] * 255 , drgb[1] * 255 ,  drgb[2] * 255 );
     } // convert
 
 protected:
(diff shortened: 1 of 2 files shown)


grimmerk (commented):
I notice that there is a piece of the code
```
m_frameCache.setMaxCost( 1 * 1024 * 1024 * 1024 );
```
should we enlarge it? @CNOCycle , @kswang1029

grimmerk:
From Susan:  The redesign issue (9) is in a part of the code that I did not write so without a lot of work, I am not qualified to comment on it.  My worry with it would be that if you are not caching things like the zoom, pan, etc in the cache id, then you might pull out a cached image when you should not.  However, I assume you have also checked that something like that does not happen.

grimmerk (commented):
This commit should not have problems, just curious that "auto-rounding is better than old way" is based on experiments/measurement or some document/theory?

carta/cpp/CartaLib/PixelPipeline/IPixelPipeline.h
```diff
@@ -34,7 +34,7 @@ namespace PixelPipeline
 ///   - can have mulitple sub-steps
 ///   - configurable
 /// - stage 5: Drgb -> QRgb
-///   - simple conversion to 8 bit values using rounding
```
grimmerk, line 37: test review

CNOCycle:
According to the C++ standard document:
`double round( double arg );`
SO, It still does the converting from double into int after calling round function.
std::round is more expensive than auto-rounding and the result may slightly different by auto-rounding if arg is very close to an integer value.
I think that it is safe in this case because we do not need high precision but high performance.

grimmerk:
test

pfederl:
From a quick look at the changes I have no issues with the proposed changes. Yes, the automatic rounding could perform better at the cost of losing some accuracy. The users would probably not notice the slight change in color.

kswang1029:
For this fix, I confirm that the performance is enhanced by a factor of few (2-4) (checked with playing a movie using the animator)

kswang1029:
Confirmed. The performance of zoom and pan is enhanced by this revised approach.

PR #10, opened 2017-01-12
fixed memory leaking when image is permuted

CNOCycle:
1. fixed memory leaking in class CCImageBase

But storing a copy of permutation of image is not an elegant method which wastes a lot of memory space when raw data is large.
My suggest is that accessing raw data by the iterator which keep the information about the permutation of image.


grimmerk:
It is ok to accept this pull request.

kswang1029:
"fixed memory leaking in class CCImageBase"
==> confirmed to work

PR #11, opened 2017-01-12
Ecchen/changed default - The default value for Clipping is 100% + resetPan when displayAxes changed

CNOCycle:
1. resetPan when displayAxes changed
when permuting axes, resetPan should be issued automatically, otherwise user will get confused where the image is.

2. The default value for Clipping is 100%

### DIFF
--- a/carta/cpp/core/Data/Image/Controller.cpp
+++ b/carta/cpp/core/Data/Image/Controller.cpp
@@ -899,7 +899,7 @@ void Controller::_initializeCallbacks(){
 void Controller::_initializeState(){
 
     //First the preference state.
-    m_state.insertValue<bool>( AUTO_CLIP, true );
+    m_state.insertValue<bool>( AUTO_CLIP, false );
     m_state.insertValue<bool>(PAN_ZOOM_ALL, true );
     m_state.insertValue<bool>( STACK_SELECT_AUTO, true );
     m_state.insertValue<double>( CLIP_VALUE_MIN, 0.0 );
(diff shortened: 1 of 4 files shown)


low-sky:
I'm not sure about the call for clipping on the 0%-100% quantiles. Was there a strong motivation for this change?  Talking to some users, the 95% default clipping had some preference.  It's certainly a good move for VLA and ALMA data.  The other change seems fine.

grimmerk:
@CNOCycle could you explain more about the effect/reason of "resetPan when displayAxes changed" to let anyone (e.g. @kswang1029 knows) the detail? So we can know how to evaluate it.

CNOCycle:
The previous version of CARTA tries to resetPan when displayAxes is changed.
It fails because some information are not updated.
And it works now.

kswang1029:
To @low-sky , thanks for sharing your thought about the default clip level for the imageViewer.

After consulting some experienced users in our institute, we feel that the default should better be 100% (no clipping applied). We understand that in some cases like high contrast images (no matter it is infrared, radio, mm), default 95% is definitely helpful than 100% for users to see low level emission. However, not every image has high contrast so in this case users will see the over-processed image. We think since the clip capability in the imageViewer is "**assistive**", specialized for high-constrast images, this capability should be triggered by users. The move to the 100% clip default is also helpful to improve performance since calculating quantiles are expensive. We better trigger the quantiles calculation when users trigger the clip, not every time when the image is opened. 

Please let us know if there is further concern about this change. Thanks again for your input.

low-sky:
Thanks.  That's a fair point.  It would be a nice thing to have as a user preference to have when we get to that point (and indeed should be preserved when someone saves a layout and tool configuration).  As long as it's just a single click to get back to 95%, it's not a big loss.

kswang1029:
We will think about how to have the user to have the freedom to choose their default clip level, together with other customizable parameters. Thanks :)

grimmer0125:
After discussion, the default check status of recompute clips on new frame will be in another ticket/issue.

grimmerk:
After discussion with @CNOCycle , we decide the default check status of recompute clips on new frames will be in another ticket/issue.

kswang1029:
"resetPan when displayAxes changed"
==> confirmed to work properly. 

The default value for Clipping is 100%
==> confirmed to work properly.

PR #12, opened 2017-01-12
fix 1. z-axis is not channel 2. crash due to {nan, nan} in quantile 3. bad_alloc problem

CNOCycle:
1. Animator can work when z-axis is not channel
2. fixed bad_alloc problem because of syntax error

N.B. we should think carefully about when function "quantiles2pixels" in quantileAlgorithms.h return {nan, nan}.
it may cause CARTA to crash because of function "convert" in IPixelPipeline.h)

### DIFF
--- a/carta/cpp/CartaLib/PixelPipeline/IPixelPipeline.h
+++ b/carta/cpp/CartaLib/PixelPipeline/IPixelPipeline.h
@@ -228,10 +228,7 @@ public:
     virtual void
     convert( double p_val, NormRgb & result ) override
     {
-        // this is a workaround solution
-        // we should think carefully about how to deal with
-        // when the upper bound and the lower bound are nan value
-        // CARTA_ASSERT( ! std::isnan( p_val ) );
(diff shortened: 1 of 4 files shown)


grimmerk:
@CNOCycle:

1. Regarding Animator can work when z-axis is not channel.  -> Could you describe the previous non-working situation? Just leave a better record. 

2.   fixed bad_alloc problem because of syntax error -> Could you tell us (actually @kswang1029 ) how to produce this issue (he will use the version before fixing this crash)

kswang1029:
"Animator can work when z-axis is not channel"
==> here I confirm that animator works with axis permutation. Good job!

"fixed bad_alloc problem because of syntax error"
==> if this issue is relevant to when all pixels are NaN, CARTA will crash, then here I confirm this is fixed. When all pixels are NaN, the color will be assigned to the lower-end of the color map. For example, if the colormap='jet', NaN pixel is colored as blue. 
Nice work ;)

PR #14, opened 2017-01-18
re-use gridline

grimmerk:
Don't calculate grid again when switching between different channels.

@kswang1029 , @CNOCycle From Susan: I guess with the re-use gridline issue (6), I would worry that the display axes would get permuted and the grid would not get redrawn, but I assume you have checked and that does not happen.

make the pull request again. https://github.com/CARTAvis/carta/pull/6 seems disappeared.


kswang1029:
Confirmed. The smoothness of the movie (animate through different channels) is significantly improved by this tweak.

PR #15, opened 2017-02-03
Fix animator drop frame

grimmerk:
Target: desktop version. (Server version share 80% of this issue so server version should be improved too but need test individually. )

Although it may not be completed yet, but create this pull request first to have a discussion space here. Please ignore some printed logs which will be removed soon.

### DIFF
--- a/carta/cpp/CartaLib/IRemoteVGView.cpp
+++ b/carta/cpp/CartaLib/IRemoteVGView.cpp
@@ -96,9 +96,14 @@ LayeredRemoteVGView::scheduleRepaint( qint64 id )
         id = m_repaintId + 1;
     }
     m_repaintId = id;
-    if ( ! m_timer-> isActive() ) {
-        m_timer-> start();
-    }
+
+    // will not enter here
+    qDebug()<<"grimmer timer1";
+    p_timerCB();
+
+//    if ( ! m_timer-> isActive() ) {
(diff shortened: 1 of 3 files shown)


kswang1029:
Hi @low-sky , for this pull request, we are trying to fix the issue that when we play movie, some frames may occasionally be skipped out, due to the original design of the CARTA animator. Next step we wish to have the fps fixed when playing movie (now you can see the fps is not stable and depends on the cpu load for other calculations like histogram/contours). Do you have any suggestions or thoughts on this issue? Thanks. :)

low-sky:
I gave this some thought and didn't have any serious concerns about targeting a fixed frame rate, especially if some pre-rendering of the different components of each frame can take place. The smoothness would be good and I agree with @kswang1029 that it's important to have smoothness for interpreting movies.  I wasn't able to view the movies on the slack channel; if it's important, can you send me the link direct on that channel.

kswang1029:
@low-sky here I attach the two movies for this issue. The test image cube is attached as well. 

[carta_movie_missing_frame.zip](https://github.com/CARTAvis/carta/files/759795/carta_movie_missing_frame.zip)

kswang1029:
I think this fix works. No obvious missing frame is spotted. Nice fix. Please consider to merge to dev. :)

PR #16, opened 2017-02-08
Ecchen/wcsplotter, drawing grid line becomes faster since skip some FITS keywords.

CNOCycle:
1. Skip FITS keywords 'ORIGIN', 'HISTORY' and 'COMMENT' so AST library processes header more efficiently.

2.  Gridline plotter with a modified header
This method works successfully on celestial coordinates, but the AST library does not support the Dec-Vel coordinate system.

3. Please let me know if some keywords which be edited are missing when FITS header is modified.


grimmerk (approved):
Regarding https://github.com/CARTAvis/carta/pull/16/commits/c544deebfb5473601005a60e5a84ab5acf92bb29, `remove unused code in AstGridPlotter.cpp`
1. The original code will handle
`permuted != true also isCelestialPlane =true`, 
But your change will not handle this case, do this case will not exist? (Initial plane is celestial?) @kswang1029 & @CNOCycl

2. Removing `_make2dFrameCelestial` is because now we use the way to change the AST header to handle different coordinate/permutation ?? (  https://github.com/CARTAvis/carta/pull/16/commits/d1a0bc17ac39cf754dc905384b34de769c4b5a0d ), @CNOCycle

CNOCycle:
1. In general, before feeding the AST library, the permutation is processed in `AstWcsGridRenderService::_getFitsHeaderforAst`

2. However, if AST dose not support Dec-Vel coordinate system(`isCelestialPlane !=true` && ( (axis1==Dec && axis2 ==Vel) ) ||  (axis2==Dec && axis1 ==Vel) ) )), we should make a coordinate transformation.

kswang1029:
After trying few test images, I think this fix works. I suggest we merge it to dev and I will keep an eye on this to see if there is missing header keyword. Good job. :)

kswang1029:
BTW, in what circumstances the axis will be parsed as Linear?

PR #17, opened 2017-02-10
Remove 'location'(or 'index') in the percentile function in DataSouce.cpp

CNOCycle:
1. Remove 'location'(or 'index') in the percentile function in DataSouce.cpp

  The first element in the `intensities` pair is the location of pixel, but this value has no physical meaning and affects the performance.

2.  Remove `divisor`

  `m_cachePercentiles` fails if 'location' is divided by 'divisor' when data is a single 2D image.

3. We should aware that `percentiles` value is stored in double type and it has limited significant digits.

  I am not sure how many significant digits are required for a cache hit in Andianna's new cache ( `percentiles` in this commit has 6 significant digits) but it should be fine if users only uses pre-defined clipping value.

### DIFF
--- a/carta/cpp/core/Data/Image/DataSource.cpp
+++ b/carta/cpp/core/Data/Image/DataSource.cpp
@@ -383,10 +383,7 @@ std::vector<std::pair<int,double> > DataSource::_getIntensityCache( int frameLow
 
         if ( allValues.size() > 0 ){
 
-            int divisor = total_size;
-            if (spectralIndex != -1) {
-                divisor /= dims[spectralIndex];
-            }
+            int divisor = 1;
 
             // only compare the intensity values; ignore the indices
             auto compareIntensityTuples = [] (const std::pair<int,double>& lhs, const std::pair<int,double>& rhs) { return lhs.second < rhs.second; };
--- a/carta/cpp/core/Data/Image/DataSource.cpp
+++ b/carta/cpp/core/Data/Image/DataSource.cpp
@@ -339,6 +339,7 @@ std::vector<std::pair<int,double> > DataSource::_getIntensityCache( int frameLow
     //See if it is in the cached percentiles first.
     int percentileCount = percentiles.size();
     std::vector<std::pair<int,double> > intensities(percentileCount,std::pair<int,double>(-1,0));
+
     //Find all the intensities we can in the cache.
     int foundCount = 0;
     for ( int i = 0; i < percentileCount; i++ ){
@@ -352,7 +353,7 @@ std::vector<std::pair<int,double> > DataSource::_getIntensityCache( int frameLow
     //Not all percentiles were in the cache.  We are going to have to look some up.
     if ( foundCount < percentileCount ){
 
-        std::vector<std::pair<int,double> > allValues;
+        std::vector<double > allValues;
         int spectralIndex = Util::getAxisIndex( m_image, AxisInfo::KnownType::SPECTRAL );
 
         Carta::Lib::NdArray::RawViewInterface* rawData = _getRawData( frameLow, frameHigh, spectralIndex );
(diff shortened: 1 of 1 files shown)


confluence:
1. The location value is used by functions in Colormap (it is stored in the state in `Colormap::_updateIntensityBounds` and retrieved and used in `Colormap::_convertIntensity`). Unless that is also changed, this part of the calculation cannot simply be removed. The function needs to be split into two functions; one which calculates the location and one which does not.  To the best of my knowledge, this is the only part of the code which uses that location value.

(Why set the location value to a dummy value based on the percentile? This is misleading -- it suggests that the value is meaningful, when it is not. The function which does not calculate the location should just return a single value instead of a pair, and the calling code should be refactored. This should also allow us to ensure that the functions calling the function which does not return a location do not use the location.)

2. The purpose of the divisor is to convert a global index within the image to a frame. That is the meaning of the location value as used by the unit conversion.  This part of the calculation should obtain the divisor by dividing the image size by the number of frames. This should result in the divisor being the size of the frame in a 3D or 2D image. In a 2D image (i.e. an image with one frame), the location will always be 0, which is correct (the index of the first frame). *I have edited this explanation; I was initially misremembering how the calculation works.* This is academic if the location is removed completely, but the location should not be removed completely.

grimmerk:
@confluence , after reading your comments and talking with @CNOCycle's, I'm a little confused. It seems that you both right but some contradiction. Sorry that I didn't trace the code and may ask you some stupid questions. 

Regarding these two commits
`That is the meaning of the location value as used by the unit conversion.` is related to `Colormap::_updateIntensityBounds`, is this correct?  

Regarding `Remove 'location'(or 'index') in the percentile function in DataSouce.cpp`
question1: 
Remove the `index part` in this function,  will improve the performance to render a fits file (less calculation), is this correct?   

question2:
From @CNOCycle's talk, if someone (not the rendering part in question1, the below b.,  unit conversion/updateIntensityBounds?) else always need the index part, so `removing index` is not possible or not meaningful. 

I think the key points in your different thoughts may be 
  a. the part which does not need `index part`  <- will always run a. when rendering 
  b. the part which does need `index part` <- will always run b. when rendering?  OR it is called/needed in seldom time?  if the answer is seldom, the whole calcuation time will become 
A-0.5 (some number less than original calculation time-1) + possible B*1(recalculate) from 1, it this correct? If so, we can use your ways to divide to two functions.

grimmerk:
By the way, we notice that histogram seems to have some `clipping` behavior, not sure if it will need `index part` or not.

confluence:
Yes, this is related to `Colormap::_updateIntensityBounds`. I believe that this is the only calling function which makes use of the location value.

1. Yes, removing this part of the calculation does improve the performance. Less memory is be used to make a copy of the data (because only the values are copied, not the values and the indices). I assume that it also takes less time to copy the data and therefore the entire calculation is faster. (I'm not sure whether the nth element search itself is also faster.)

2. I'm not sure if I understand the question. The proposed change (removing the location calculation entirely) breaks `Colormap`, because `Colormap` needs the location. It is necessary to retain the option of finding both the value and the location. However, it is not necessary most of the time, so it is a waste to find both values most of the time.

I previously considered having only one function which finds the value, and optionally looking up the location afterwards in a separate step, but this is slower than the current solution of performing nth-element on the tuples. That is why the current solution is in place.

You could keep a single function and pass in a boolean parameter which determines whether the location is calculated or not -- but then the return type would have to be the same in both cases, and then in 95% of cases you would be returning pairs with unused dummy values, which seems untidy and confusing.

For this reason, I suggest the solution of two functions: `getIntensity`, which returns a vector of single intensity values, and `getLocationAndIntensity`, which returns a vector of (location, value) pairs.

I think that it makes sense for the function which only finds the intensity, which is the most common case, to be called `getIntensity` (I mean that the public "getter" is called that; that's a wrapper around `_getIntensityCache`. I'm not convinced that it's necessary to maintain these multiple layers; maybe it would be simpler to collapse them) -- so this pull request is part of the way there. I suggest the following additions:

1. Add a second function, which is called something like `getLocationAndIntensity`, and restores the location + value calculation which was removed from `getIntensity`. This also needs to undo the change to the divisor calculation, because the changed version was incorrect. All the necessary wrappers need to be added for this new function (in `DataSource`, `LayerData`, `LayerGroup`, `Controller` -- all the places where `getIntensity` is currently defined).

2. Make `Colormap` call `getLocationAndIntensity` instead of `getIntensity`.

3. Stop returning pairs from `getIntensity`; return just the intensity values instead. Update all the calling functions. This change should unambiguously guarantee that everything calling `getIntensity` does not use the location, although I understand that this is a tedious bit of refactoring. This is not strictly necessary to do in the first pass and could be done later.

It may be possible to use polymorphism in some way to factor out some of the common parts of the two functions, so that code which is conceptually the same is not duplicated, but that is not absolutely necessary (and I can attempt to do this when I merge in my implementation of the approximate quantile calculation).

If you don't want to change the return value of the function at this stage, you could just use the solution with one function and a boolean parameter (which defaults to false, and is set to true only by `Colormap`).

grimmerk:
If adapting polymorphism way, the return type should be the same (pair) so that one will include dummy (index) value?  But it seems that the needed change in this way is the least one.

confluence:
Yes, leaving the same return type is the least disruptive way to change this, and it could be the first step to a more extensive change.

If you want to do this, then it will be even simpler to keep one function and add a boolean parameter instead of making two functions.

I would strongly suggest setting the dummy value to e.g. -1 to make it absolutely clear that it's a meaningless value, instead of setting it to arbitrary varying values.

kswang1029:
Hi @confluence (cc @grimmer0125 @CNOCycle ) regarding this "index" issue, we noticed a bug in the colormap plugin. To my understanding, index is used to derived frequency information and frequency can be used for unit conversion (Jy/beam <-> Kelvin). In the colormap plugin, the default unit as seen in GUI is ALWAYS Jy/beam regardless the unit defined in the image header. For example, orion_12co_hera.fits is in unit of Kelvin (as seen in the cursor information at the bottom of the imageViewer). In this case, colormap plugin still use Jy/beam as the unit. The bug here is that if the user switch the unit from Jy/beam to Kelvin, the conversion still proceeds and the result (the lower and upper bounds) is not correct. Whether we decide to remove index or not, or create something equivalent to index (either two-function approach or polymorphism), we should keep the unit conversion issue in mind.

grimmerk:
Done in #200 .

PR #19, opened 2017-02-24
CAS-9784 fix for Loading CasaImage

CNOCycle:
1.  fix for Loading CasaImage problem (casa image file, casa image format)
2. add `_getPermOrder()` in `DataSource` class

### DIFF
--- a/carta/cpp/core/Data/Image/DataSource.cpp
+++ b/carta/cpp/core/Data/Image/DataSource.cpp
@@ -71,6 +71,24 @@ DataSource::DataSource() :
         }
 }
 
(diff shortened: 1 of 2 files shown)


grimmerk:
More detail from @kswang1029 : 

Back to the issue of opening casa image format, the root cause is the axis order of stoke and channel. FITS standard put channel and stoke to the 3rd and 4th axis, respectively, while casa format puts stoke to the 3rd axis and channel the 4th. Therefore, carta somehow uses wrong information so that when you change "channels" to index out of the array length of the stoke axis, no image will be shown.

PR #21, opened 2017-03-03
Known issue:  encoding format for FITS header

CNOCycle:
1. There are many encoding format for FITS header
ref: http://www.starlink.rl.ac.uk/docs/sun210.htx/sun210se17.html

2. Here is the FITS standard v4.0
ref: https://fits.gsfc.nasa.gov/standard40/fits_standard40draft1.pdf

3. CasaImage does not follow FITS-WCS standard, so I'm trying to fix it
CasaImage format    -> PC%2d_%2d
but,
FITS-WCS standard -> PC%1d_%1d

### DIFF
--- a/carta/cpp/plugins/WcsPlotter/AstWcsGridRenderService.cpp
+++ b/carta/cpp/plugins/WcsPlotter/AstWcsGridRenderService.cpp
@@ -171,6 +171,10 @@ QString AstWcsGridRenderService::_getFitsHeaderforAst(QStringList &fitsHeader)
     lambdareplace1D("PV%1_3" );
     lambdareplace1D("PV%1_4" );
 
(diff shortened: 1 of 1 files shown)


CNOCycle:
This pull request should be closed since `_getFitsHeaderforAst(QStringList &fitsHeader)` function is removed and this patch move to `FitsHeaderExtractor` class after a fix for CAS-9839 inaccurate gridline plotting.
FITS header are parsed with casacore because we are not able to test all kinds of FITS formats one by one.
But casacore is broken when image is rotated and its PC matrix is not a identity matrix.
These issues with PC matrix has been reported to NRAO. Hope they will fix problems in the short time.

grimmerk:
What do PC%1d_%1d & PC%2d_%2d refer?

PR #22, opened 2017-03-03
Fix for Statistics plugin to work on both Mac and Linux

grimmerk:


grimmerk:
It is solved by this commit, https://github.com/CARTAvis/carta/commit/76d1afea34c388d84d6250ce839f7f9302f59cc2, so just close.

PR #23, opened 2017-03-05
refine displayed cursor info and add more precision to displayed coordinates

kswang1029:
1) refine the displayed cursor information for better readability 

2) add more precision to the displayed coordinates, including frequency axis. This is needed for high-resolution observations like HLtau.

Attached image is
screen shot 2017-03-05 at 6 41 39 pm
 before and after comparison.

### DIFF
--- a/carta/cpp/core/Data/Image/DataSource.cpp
+++ b/carta/cpp/core/Data/Image/DataSource.cpp
@@ -194,11 +194,11 @@ QString DataSource::_getCursorText( int mouseX, int mouseY,
 
         QString pixelValue = _getPixelValue( round(imgX), round(imgY), frames );
         QString pixelUnits = _getPixelUnits();
-        out << pixelValue << " " << pixelUnits;
-        out <<"Pixel:" << imgX << "," << imgY << "\n";
+        out << "Pixel value:" << pixelValue << " " << pixelUnits << "at ";
+        out <<"pixel:" << imgX << ", " << imgY << "\n";
 
         cf-> setSkyCS( cs );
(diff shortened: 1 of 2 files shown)


low-sky:
👍  
I was thinking about this and wondered about two things.

First, in RA/Dec coordinates, typically the RA has one additional decimal place relative to the declination since hours of arc are larger an order of magnitude larger than degrees.

Second, should the precision that is displayed depend on either (1) the image size shown in the window or (2) the CD values of transform.  It's great to have high precision for milliarcsecond pixels, but it's also a bad idea to show this much precision when your pixels are arc-minute scale.

kswang1029:
Thanks for the comments. I noticed that the display precision should better be determined by the pixel size and/or zoom level. Same issue in the statistics plugin as well. For dynamic precision it takes more efforts to accomplish, while this simple fix is done by simply modifying the precision defined in the code and perhaps useful for the pre-release version. I am fine to withdraw this pr and later fix all precision issue with a new one.

low-sky:
I agree that dynamically changing the precision is difficult with state of the viewer.  The ideal, intermediate solution may be to set the precision based on the pixel scale in the image at this point and leave the truly dynamic precision for later.  If that's also quite difficult at this point it would be still good to get this change in.

kswang1029:
ok I have modified the displayed coordinate again to show less digits. Now it is .xxxx. This should be already more than casaViewer which shows .xxx
screen shot 2017-03-06 at 1 02 49 pm
screen shot 2017-03-06 at 1 03 08 pm

PR #25, opened 2017-03-06
show beam size with dynamical precision

kswang1029:
modified the displayed formats of beam size in statistics plugin.

Now the precision is dynamically determined by the header beam info. Proper unit (arcsec or arcmin) is dynamically displayed too.

N.B. the degree symbol needs to be worked out still, likely due to encoding issue. I can correctly display the degree symbol (check my comment-out code) but extra character is also shown...

screen shot 2017-03-06 at 5 40 50 pm


kswang1029:
screen shot 2017-03-07 at 1 07 56 pm
regarding "f4bd9aa", this fix doesn't work. This fix gave same result as my comment-out code. I think we can just use "deg" for the moment (my commit) and later when we have time to investigate this issue.

PR #29, opened 2017-03-10
kswang/fix colormap reverse ui

kswang1029:
fixed the issue that when "reverse" is toggled, colorbar isn't reversed.


carta/html5/common/skel/source/class/skel/widgets/Colormap/ColorGradient.js
```diff
@@ -94,7 +94,7 @@ qx.Class.define("skel.widgets.Colormap.ColorGradient", {
          * @param invertMap {boolean} true to invert the map; false, otherwise.
          */
         setInvert : function( invertMap ){
-           if ( invertMap != this.m_invert ){
```
grimmerk: I don't know why the original expression did not work. When we change people, is it that some place will 
1. change this.m_invert
2. then call this function
3. so the original condition does not work. <- is 1+ 2 the reasons?

grimmerk (approved):

PR #32, opened 2017-03-13
Fix the bug that Carta will pop up a error window when people close Carta program. 

grimmerk:


carta/cpp/core/Viewer.cpp
```diff
@@ -123,7 +124,24 @@ Viewer::start()
...
+    if ( res.isNull() || ! res.val() ) {
+        qWarning( "Could not find a disk cache plugin." );
+        m_diskCache = nullptr;
+    }
+    else {
+        m_diskCache = res.val();
+        qDebug() << "Check the number of owners:" << m_diskCache.use_count();
+    }
+
+    m_diskCache->Release();
```
grimmerk: m_diskCache->Release(); -> should move to 
else {
}
?

carta/cpp/core/Viewer.h
```diff
@@ -41,6 +41,10 @@ class Viewer : public QObject
 
 public slots:
 
+    /// Closing the database of cache before quit
+    /// It had better be executed by other object (such as any unloaded plugin when exitting)
+    void DBClose();
+
```
grimmerk, line 47: test

PR #39, opened 2017-03-27
Solve Animator crash.  

grimmerk:
This crash bug is the side effect of the commit "9da880b911066b0c2c1b207588f84d29cf7a632c" (Fix animator drop frame). It would let animator happen crash when opening two datasets.


kswang1029:
here I confirm that now carta won't crash when playing movie with two image cubes that have different numbers of channels opened.
The channel indices are sync'd in the two cubes. That means, if I stop the movie at channel 10 of cube A then switch to cube B, I will see channel 10 as well. If I stopped at the channel number that the other cube doesn't have, I will see the last channel.

PR #41, opened 2017-03-27
Cjhsu/fix image load

appolloford:
Based on the content of files/folders to set the icons of them in the tree.


grimmerk:
@kswang1029 can you build and test well?  circleci shows something wrong but it may not be meanful.

kswang1029:
This branch cannot be built with docker due to fitter1d.pro. Using fitter1d.pro from develop branch can solve this. 

I have tested this branch and it seems to work well. However, there is one case that still can cause mis-identification of file type. To reproduce this issue, I put a file ended with xxx.fits.header with only the header information in the file. See attachment. Perhaps we can use *.fits && first line contains "SIMPLE  =                    T /Standard FITS " to better identify FITS image? 
[hack.zip](https://github.com/CARTAvis/carta/files/895158/hack.zip)

PR #46, opened 2017-03-29
Mark/re locate config json

markccchiang:
The main changes are:
1. Add a main json configure file in the directory: $CARTAWORKHOME/CARTAvis/carta/config.json
2. Modify the files "CmdLine.h" and "CmdLine.cpp" to load "config.json" in the item 1 directory.
3. This branch is merged with the branches "grimmer/fixIncompatibleAxesCrash" and "grimmer/fixAnimatorCrash"


CNOCycle:
I think that `config.json` should be copied from source directory to build/installed directory during the building stage.

Otherwise, this commit may fail when the developers use out-of-source build or CARTA is installed in other places or the source code has been removed.

for example, my file structure is
```
├── build
│   ├── debug
│   └── release
└── carta-src
```

markccchiang:
This is a good point. So we may modify "carta.pro" to copy the file
"config.json" from source directory to the root of build directory. The
search directory setting in "CmdLine.cpp" also need to change. That is my
current idea.

2017-03-30 16:18 GMT+08:00 CNOCycle <notifications@github.com>:

grimmerk:
https://github.com/CARTAvis/carta/pull/57 is replacing this. so close it

PR #49, opened 2017-03-30
Fix the problem of CachePercentile

appolloford:
The judgment of LeastRecentlyUsedCacheEntry is not complete so that carta can't distinguish the difference between the entries of maximum and minimum. This causes the entry of maximum won't be saved and read later. After adding these conditions, I think the entries can be saved correctly now.


grimmerk (commented):
From the code of `m_cachedPercentiles.put( frameLow, frameHigh, intensities[i].first, percentiles[i], intensities[i].second );` and its below code, I guess, `m_intensity == other.m_intensity` may be unnecessary? Since it looks like it is the value/target part,  not the key part, could you confirm with @CNOCycle

appolloford:
Yes, I also hesitated to add this condition. However, I still added this one because it may be helpful to avoid some accident. I will discuss with @CNOCycle on the next wednesday.

grimmerk (approved):

PR #54, opened 2017-03-31
Mark / create "make install" 

markccchiang:
This branch is in response to the issue raised by CNOCycle, where I merge and improve previous pull requested branches: "mark/addGridLinesOption", mark/removeGridLines", "mark/reLocateConfigJson", "grimmer/fixIncompatibleAxesCrash" and "grimmer/fixAnimatorCrash". The changed files compared to my latest pull request are 

$CARTAWORKHOME/CARTAvis/carta/config/config.json
$CARTAWORKHOME/CARTAvis/carta/carta.pro  and  
$CARTAWORKHOME/CARTAvis/carta/cpp/core/CmdLine.cpp

We have used "make -j2" to build the Carta viewer, now we can use "make install" command to copy the necessary files or data in the building directory. For example, in my case I copy the main configuration folder and file "config/config.json" to the root of building directory. So users can built their own Carta viewer in any directory without been affected by the absolute path that set in the Carta code.

### DIFF
--- a/carta/carta.pro
+++ b/carta/carta.pro
@@ -14,11 +14,18 @@ dbg( "===---------------------------------------------====")
 
 TEMPLATE = subdirs
 SUBDIRS += cpp
-SUBDIRS += scriptedClient
-SUBDIRS += scripts
+# SUBDIRS += scriptedClient
+# SUBDIRS += scripts
 
 OTHER_FILES += readme.txt uncrustify.cfg
 
+#
+# Apply make install option to copy the config folder and its file to the build directory
+#
+install_config.path = $${OUT_PWD}
+install_config.files += config/
+
+INSTALLS += install_config
 
 # make sure user did not specify debug version...
 # it's probably harmless, but that's probably not what the user wanted
--- a/carta/config/config.json
+++ b/carta/config/config.json
@@ -1,0 +1,15 @@
+{
+    "_comment" : "List of plugin directories",
+    "pluginDirs": [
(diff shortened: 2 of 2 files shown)


grimmerk:
@markccchiang , for developing purpose, manually typing `make install` seems not straightforward, and not simplified. My understanding is, usually it is to install the program to other places. If we put this in Qt creator, it will redundantly copy. Or does `make install` is only for packaging, in your design?

markccchiang:
"make install" copys the build execution and necessary files/codes to the
specific directory we set. It is for packaging from this aspect. In this
case, I just want to make sure the execution program can always find the
"config.json." We don't need to use "make install" once the Carta build
directory is established. I have set the priority search path of
"config.json" in users' home directory, users can change the main
configuration under the home directory, another "config.json" under the
Carta build directory is only for backup.

2017-04-06 10:58 GMT+08:00 Grimmer <notifications@github.com>:

grimmerk:
I think moving this action into qt is the better way. But it may be ok to still use the current way, as long as 
1. rename make install to others, e.g. make install_config
2. if the file exists, does not copy.

markccchiang:
Ok. I have create another pull request #57 which moves this action into qt
"make".

2017-04-06 15:39 GMT+08:00 Grimmer <notifications@github.com>:

grimmerk:
Since https://github.com/CARTAvis/carta/pull/57 is the replacement of this pull request, so just close it.

PR #58, opened 2017-04-10
fix for re-use gridline

CNOCycle:
In pull request #14, #6, it may not get the right result.
Gridline can be reused only when switching channels

### DIFF
--- a/carta/cpp/plugins/WcsPlotter/AstWcsGridRenderService.cpp
+++ b/carta/cpp/plugins/WcsPlotter/AstWcsGridRenderService.cpp
@@ -501,7 +501,10 @@ void AstWcsGridRenderService::setAxisDisplayInfo( std::vector<Carta::Lib::AxisDi
         for ( int i = 0; i < infoCount; i++ ){
             if ( displayInfos[i] != m_axisDisplayInfos[i] ){
                 m_axisDisplayInfos[i] = displayInfos[i];
-                //m_vgValid = false;
+                if(displayInfos[i].getAxisType() != Carta::Lib::AxisInfo::KnownType::SPECTRAL)
+                {
+                    m_vgValid = false;
+                }
             }
         }
     }


carta/cpp/plugins/WcsPlotter/AstWcsGridRenderService.cpp
```diff
@@ -501,7 +501,10 @@ void AstWcsGridRenderService::setAxisDisplayInfo( std::vector<Carta::Lib::AxisDi
         for ( int i = 0; i < infoCount; i++ ){
             if ( displayInfos[i] != m_axisDisplayInfos[i] ){
                 m_axisDisplayInfos[i] = displayInfos[i];
-                //m_vgValid = false;
+                if(displayInfos[i].getAxisType() != Carta::Lib::AxisInfo::KnownType::SPECTRAL)
```
grimmerk, line 504: It seems that ra, dec, stokes will not re-use grid line but may be OK, stokes will not have too many z values.

kswang1029: to have consistent logic, we should apply reuse when the animated axis is stokes, same as channel.

kswang1029:
I have made an image cube with wide field to test this PR. Ideally we should see gridline is recalculated when viewing RA v.s. channel

Will update my test result soon.

kswang1029:
To update, I tried to load my wideField image cube to casaViewer and tried to permute the axes to 1) RA vs channel and 2) Dec vs channel. In both cases, I do see the gridlines belonging to RA or Dec updated per frame. The image cube can be found at acdc1:/data/carta_image_library/Images/wideFieldCube.fits

kswang1029:
Update: This PR doesn't fix the issue that when image cube is permuted, for example, to RA vs channel, the gridline per Dec frame should be all different. It is likely the gridline stays at Dec_frame=0.
screen shot 2017-04-20 at 3 20 51 pm

PR #81, opened 2017-04-27
File list in animator

grimmerk:
Previous:
Used a textField to show the current image and switch it to switch in animators. But no file name information. Only index part. 

Now:
Use a combobox to switch and include the file name information of each file. 

There are two ways to add file list in animators. 
1. Send stack if to animator related UI (e.g. animator.js or DisplayWindowAnimator.js), then they can use this id to get the stack information of the related images which having file list. But the problem is that DisplayWindowImage.js parse the stack info and store inside. but not store somewhere others can access. So other objects need to use this stack id to get the raw string from DesktopConnector.js ( or server connector), then parse again. 

2. The current solution, Controller.cpp sends data into Animator.cpp , AnimatorType.cpp, Selection.cpp , then flush the file list info to Animator.js. (e.g. Animator.js is related to AnimatorType.cpp). 

p.s. 
this issue uses qx.ui.form.ComboBox, not qx.ui.form.SelectBox. It is because the latter one will not dismiss automatically.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ build*/
 PureWeb*.lic
 
 carta.pro.user
+node_modules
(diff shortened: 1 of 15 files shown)


kswang1029:
I tested a bit with this PR. No obvious issue is found. I also tested by opening 11 files to see if it works or not, and the answer is yes. 
If possible, perhaps we can make the drop-down menu wider in order to show more characters of file name, since there is still plenty of space in the same row.

Issue #87, opened 2017-04-28
UI not sync: Clip related. Some are fixed. The remaining: clipping text will not be updated when switching to 2nd file 


grimmerk:
1. [x] colormap 沒有同步, quantile有作用時
the upperbound and lowerbound are not updated if AUTO_CLIP is true?
==> if "recompute clips on new frame" is enabled, the upperbound and lowerbound in colormap should be updated when switching to different frames, since upperbound and lowerbound are determined by the pixel values of the CURRENT frame, not the whole cube.

update: this is workaround fixed in https://github.com/CARTAvis/carta/pull/115#issuecomment-305374358. also #162 

2. [ ] clipping百分比沒有同步: 切到第二個檔案, ui上面是圖有用到舊的clpping, 但還是show default 100%
==> when two images are opened and we applied 95% clip to image A, we should only see 95% clip is applied to image A. But when switching to image B, we see 95% clip is ALSO applied to image B. This is not the expected behavior. Clip level should be independently applied to different images and the Clip level UI should be highlighted correctly for different images. 

3. When operating clipping or colorbar on dataset-1, it always apply on dataset-2.  "Apply all" default unchecked
==> same as 2.

4. in clip-per-frame mode (quantile mode, not percentile mode), extreme values of different channels do not update? <-what meaning?
==> same as 1.

We will revise to English later after we try to clarify the reproduce step for each issue.

PR #115, opened 2017-05-03
Calculate the percentile only for Intensity stoke by default and add UI to switch stoke

markccchiang:
Grimmer update:
- [x] May need to disable switching to stoke[>=1] in animators. 
- [x] handle quantile mode (per frame mode)
---

Changes in this branch:

I modify `DataSource.h` and `DataSource.cpp` codes, add a private function `_getRawDataForIntensity(frameLow, frameHigh, spectralIndex, stokeIndex)` which is applied in the function `_getIntensityCache(frameLow, frameHigh, percentiles)`. It allows us to calculate the percentile only for Intensity if we have image files (Casa or fits/miriad image files) with multiple stokes. 

Notes:

Since the memory cache can store old information from image files. In order to test codes, we have to prepare the same image file with different names, like `HH211_IQU.image.pbcor` and `HH211_IQU_2.image.pbcor`. Each image file is used to test the codes before and after modifications.

Some Issues:

1. The dump values of intensity from CARTA is not the same with the real intensity values shown in CASA. Is there any unit transformation we need to apply first? It might affect the colormap setting.

2. The way we calculate the percentile (e.q. 99%) is to use the selection algorithm supplied by cpp std utilies. In the process, it may do some sort-like action for the data array.

The selection algorithm we use is `std::nth_element(vector.begin(), vector.begin() + nth-Index, vector.end());` one can refer to http://en.cppreference.com/w/cpp/algorithm/nth_element or http://www.cplusplus.com/reference/algorithm/nth_element/

3. Now we calculate the percentile only for Intensity stoke, but it is for all channels. Do we also need to calculate the percentile for each channel?

### DIFF
--- a/carta/cpp/core/Algorithms/quantileAlgorithms.h
+++ b/carta/cpp/core/Algorithms/quantileAlgorithms.h
@@ -55,7 +55,8 @@ quantiles2pixels(
     std::vector < Scalar > allValues;
     view.forEach(
         [& allValues] ( const Scalar & val ) {
-            if ( ! std::isnan( val ) ) {
+            // check if the value from raw data is finite
+            if ( std::isfinite( val ) ) {
                 allValues.push_back( val );
             }
         }
@@ -87,10 +88,13 @@ quantiles2pixels(
                 if( inp <= v) cnt ++;
             }
             double qq = double(cnt)/allValues.size();
-            qDebug() << "  " << q << "->" << v << qq << fabs(q-qq)
+            qDebug() << "  Set percentile" << q << "->"
+                     << "the intensity is" << v
+                     << ", its actual percentile is" << qq
+                     << ", the absolute difference between set and actual percentiles is" << fabs(q-qq)
                      << ((fabs(q-qq) > 0.01) ? "!!!" : "");
         }
-        qDebug() << "-----------------------------";
+        qDebug() << "--------------------------------------------";
     }
 
     return result;
--- a/carta/cpp/core/Algorithms/quantileAlgorithms.h
+++ b/carta/cpp/core/Algorithms/quantileAlgorithms.h
@@ -56,7 +56,9 @@ quantiles2pixels(
     view.forEach(
         [& allValues] ( const Scalar & val ) {
             // check if the value from raw data is finite
-            if ( std::isfinite( val ) ) {
+            //if ( std::isfinite( val ) ) {
+            // check if the raw data is NaN
+            if ( ! std::isnan( val ) ) {
                 allValues.push_back( val );
             }
         }
--- a/carta/cpp/core/Data/Image/Controller.cpp
+++ b/carta/cpp/core/Data/Image/Controller.cpp
@@ -1089,7 +1089,7 @@ void Controller::saveImageResultCB( bool result ){
 
 void Controller::setAutoClip( bool autoClip ){
     bool oldAutoClip = m_state.getValue<bool>(AUTO_CLIP );
-    if ( autoClip != oldAutoClip ){
+    if ( autoClip != oldAutoClip ) {
         m_state.setValue<bool>( AUTO_CLIP, autoClip );
         m_state.flushState();
     }
@@ -1335,7 +1335,13 @@ void Controller::_updateCursorText(bool notifyClients ){
     QString formattedCursor;
     int mouseX = m_stateMouse.getValue<int>(ImageView::MOUSE_X );
     int mouseY = m_stateMouse.getValue<int>(ImageView::MOUSE_Y );
-    QString cursorText = m_stack->_getCursorText( mouseX, mouseY);
+
+    // get Quantile information and show them on the image viewer
+    double minPercent = m_state.getValue<double>(CLIP_VALUE_MIN);
+    double maxPercent = m_state.getValue<double>(CLIP_VALUE_MAX);
+    bool isAutoClip = m_state.getValue<bool>(AUTO_CLIP);
+    QString cursorText = m_stack->_getCursorText(isAutoClip, minPercent, maxPercent, mouseX, mouseY);
+
     if ( !cursorText.isEmpty()){
         if ( cursorText != m_stateMouse.getValue<QString>(CURSOR)){
             m_stateMouse.setValue<QString>( CURSOR, cursorText );
--- a/carta/cpp/core/Data/Image/Controller.cpp
+++ b/carta/cpp/core/Data/Image/Controller.cpp
@@ -361,7 +361,9 @@ std::vector<std::pair<int,double> > Controller::getIntensity( const std::vector<
 }
 
 std::vector<std::pair<int,double> > Controller::getIntensity( int frameLow, int frameHigh, const std::vector<double>& percentiles ) const{
-    std::vector<std::pair<int,double> > intensities = m_stack->_getIntensity( frameLow, frameHigh, percentiles );
+    int stokeFrame = getFrame(AxisInfo::KnownType::STOKES);
+    qDebug() << "++++++++ get the stoke frame=" << stokeFrame << "( -1: no stoke, 0: stoke I, 1: stoke Q, 2: stoke U, 3: stoke V)";
+    std::vector<std::pair<int,double> > intensities = m_stack->_getIntensity( frameLow, frameHigh, percentiles, stokeFrame );
     return intensities;
 }
 
--- a/carta/cpp/core/Data/Image/DataSource.cpp
+++ b/carta/cpp/core/Data/Image/DataSource.cpp
@@ -190,7 +190,7 @@ QStringList DataSource::_getCoordinates( double x, double y,
     return list;
 }
 
-
+// print the pixel value and x-y coordinate for the cursor on the image viewer
 QString DataSource::_getCursorText( int mouseX, int mouseY,
         Carta::Lib::KnownSkyCS cs, const std::vector<int>& frames,
         double zoom, const QPointF& pan, const QSize& outputSize ){
@@ -206,14 +206,13 @@ QString DataSource::_getCursorText( int mouseX, int mouseY,
         CoordinateFormatterInterface::SharedPtr cf(
                 m_image-> metaData()-> coordinateFormatter()-> clone() );
 
-
         QString coordName = m_coords->getName( cf->skyCS() );
         //out << "Default sky cs:" << coordName << "\n";
 
         QString pixelValue = _getPixelValue( round(imgX), round(imgY), frames );
         QString pixelUnits = _getPixelUnits();
-        out << "Pixel value:" << pixelValue << " " << pixelUnits << " at ";
-        out <<"pixel:" << imgX << "," << imgY << "\n";
+        out << "Pixel value: " << pixelValue << " " << pixelUnits << " at ";
+        out <<"pixel coordinate: X=" << imgX << ", Y=" << imgY << "\n";
 
         cf-> setSkyCS( cs );
         out << "[ " << m_coords->getName( cs ) << " ] ";
@@ -372,8 +371,7 @@ std::shared_ptr<Carta::Core::ImageRenderService::Service> DataSource::_getRender
     return m_renderService;
 }
 
-// TODO: create another function which only looks for the intensity. Most calling functions don't need the location.
-// 2017/05/16    C.C. Chiang: Modify this function that can get the intensity (pixel) for different stokes (I, Q, U and V)
+// 2017/05/16    C.C. Chiang: Modify this function that it can get the intensity (pixel) for different stokes (I, Q, U and V)
 std::vector<std::pair<int,double> > DataSource::_getIntensityCache( int frameLow, int frameHigh,
         const std::vector<double>& percentiles, int stokeFrame ){
 
@@ -447,7 +445,7 @@ std::vector<std::pair<int,double> > DataSource::_getIntensityCache( int frameLow
 
         // get raw data (only for the stoke I) in order to sort the raw data set by selection algorithm
         int stokeSliceIndex = stokeFrame; // -1 is for no stoke; 0 is for stoke I; 1 is for stoke Q; 2 is for stoke U; 3 is for stoke V
-        Carta::Lib::NdArray::RawViewInterface* rawData = _getRawDataForIntensity( frameLow, frameHigh, spectralIndex, stokeIndex, stokeSliceIndex );
+        Carta::Lib::NdArray::RawViewInterface* rawData = _getRawDataForStoke( frameLow, frameHigh, spectralIndex, stokeIndex, stokeSliceIndex );
         qDebug() << "++++++++ frameLow=" << frameLow << ", frameHigh=" << frameHigh;
         qDebug() << "++++++++ set the Stoke Index for Percentile=" << stokeSliceIndex
                  << "(-1: no stoke, 0: stoke I, 1: stoke Q, 2: stoke U, 3: stoke V)";
@@ -477,6 +475,7 @@ std::vector<std::pair<int,double> > DataSource::_getIntensityCache( int frameLow
             index++;
         }
         );
+        qDebug() << "++++++++ raw data index number=" << index;
 
         if ( allValues.size() > 0 ){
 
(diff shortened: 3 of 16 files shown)


kswang1029:
Some Issues:

The dump values of intensity from CARTA is not the same with the real intensity values shown in CASA. Is there any unit transformation we need to apply first? It might affect the colormap setting.
**==> we should see the original pixel max/min values in colormap. no unit transformation should be triggered by default.** 

Now we calculate the percentile only for Intensity stoke, but it is for all channels. Do we also need to calculate the percentile for each channel?
**==> when "recompute clips on new frame" is enabled, we should search for max/min on the current frame, then pass to colormap. In this case, the min/max is a function of channel (if we are viewing ra vs dec and go through different channel). In general, min/max is a function of the animated axis (so it can be ra or dec).**

grimmerk:
@kswang1029 , @markccchiang , 

Some things:
1. When enable `recompute clips on new frame`, it will calculate `quantile` and not use percentile. So we need to check this point.

2. Even the percentile or quantile is correct (only use stoke[0]-Intensity), we need to check if the rendering only uses the pixel value at (x,y, some channel, stoke[0]) or PV image (x, channel, some-y, stoke[0]).

CNOCycle:
Please note that `std::nth_element` is `selection algorithm` instead of `sorting algorithm`.
They are totally different terminology in computer science.
ref: https://en.wikipedia.org/wiki/Selection_algorithm
ref: https://en.wikipedia.org/wiki/Sorting_algorithm

grimmerk:
@CNOCycle 
Thank you.  I have revised the sentence.

markccchiang:
### Update:

1. Solve the potential crash problem while running Quantile mode (crash example: open image files `HH211_I.image.pbcor` or `HH211_IQU.image.pbcor`, and run quantile mode).
   --> modify `quantileAlgorithms.h` 
update: this is because the file has some +-INF value but it will not happen in normal files. So not this changed is reverted .

2. Now we can run quantile mode for specific Channel or Stoke 
   --> modify `DataSource.cpp` and `DataSource.h`

3. By default, the percentile mode is only calculated for Stoke I. Now it is available for the other Stokes, but we have to add this optional item in UI.

### Todo:

1. [x] Add optional items in UI that allow us to calculate the percentile with different Stokes.

2. Solve the statistical region which is not updated while running animation (issue #142).

kswang1029:
For todo1, can we simply use the index of stokes from animator? When loading images with multiple stokes, stokes-animator should show up in the animator panel. Therefore we can use the index and pass to quantile calculator. In this way I think we don't need to add new UI. BTW, for ALMA polarization data, the stokes sequence is usually 0:I, 1:Q and 2:U. But for others, the sequence may not be identical to ALMA convention. Therefore I think using stokes index is good enough. We don't have to always assume index 0 is mapped to stokes I.

markccchiang:
Ok, I will try this way.

2017-05-11 23:05 GMT+08:00 kswang1029 <notifications@github.com>:

kswang1029 (commented):
Could you attach a screenshot for the new UI? Percentile calculations should be applied to current stokes. Could you explain why we need a new UI for percentile in animator?

markccchiang:
I will show how to use it later. Seems there is a cache bug in the current modification, I am looking to fix it first..

markccchiang:
### Update: 

Add an optional function in UI (animation panel) that users can choose different Stokes for Percentile calculation.

### Usage:

As shown in the following figure:



**Step 1**: set the stoke index on animation panel for calculating Percentile (0: stoke I, 1: stoke Q, and 2: stoke U)

**Step 2**: click different percentage buttons and check if Max and Min pixel values on the color map changed. Usually by clicking one or two different percentage buttons the cache will be refreshed for the new Stoke (this is not good and can be improved in future).

kswang1029:
ok I see. Thanks. I thought there is a "new" UI in animator. So if I use animator to switch to stokes=1 (in this case Q) and hit 95%, CARTA will use all five channels (and all pixels per channels, unless "recompute clips on new frames" is toggled) at stokes=1 to look for values at 2.5%th and 97.5%th then applies these two values to colormap? If so, that is the correct behavior we needed.

markccchiang:
Correct! I have checked this function with image files `HH211_IQU.image.pbcor`, `HH211_I.image.pbcor`, `HH211_Q.image.pbcor` and `HH211_U.image.pbcor`. Perhaps you can do more checks and let me know if there are problems.

kswang1029:
Sure I will build your branch now

kswang1029:
I spot one potential issue but need your confirmation. 
I made a new set of images based on HH211_IQU.image.pbcor and put them in our library.
HH211_IQU_zoom_4ch.image.pbcor
#          -- maximum value [max]:                    0.0664254 Jy/beam
#          -- minimum value [min]:                    -0.00370477 Jy/beam
(using all stokes and all channels)

HH211_I_zoom_4ch.image.pbcor
#          -- maximum value [max]:                    0.0664254 Jy/beam
#          -- minimum value [min]:                    -0.00370477 Jy/beam
(using stokes=I and all channels)

HH211_Q_zoom_4ch.image.pbcor
#          -- maximum value [max]:                    0.00252461 Jy/beam
#          -- minimum value [min]:                    -0.00225548 Jy/beam
(using stokes=Q and all channels)

HH211_U_zoom_4ch.image.pbcor
#          -- maximum value [max]:                    0.0022621 Jy/beam
#          -- minimum value [min]:                    -0.00238701 Jy/beam
(using stokes=U and all channels)

(I made the image smaller to get rid of the high pixels due to primary beam correction. I cut out channel=0 too)

When we open HH211_IQU_zoom_4ch.image.pbcor with clip=100% (default), we see lowerbound and upperbound are -0.00370477 and 0.0664254, respectively. Then switch to stokes=1(Q), we should see these two values are -0.00225548 and 0.00252461, respectively, but we still see -0.00370477 and 0.0664254. Please confirm if my description is true, if so, this should be corrected as well.

kswang1029:
Another test: 
1) open HH211_IQU_zoom_4ch.image.pbcor and switch to stokes=1 (Q) and apply 95% clip
we see -0.00132747 and 0.00141453 in colormap. 
2) reopen CARTA and open HH211_Q_zoom_4ch.image.pbcor and apply 95% clip
we see -0.00104755 and 0.00104178 in colormap.

In principle we should see same results for these two approaches. What do you think of this?

kswang1029:
I made a new set of images for the test and put them in our library.
HH211_IQU_zoom_4ch.image.pbcor
          -- maximum value [max]:                    0.0664254 Jy/beam
          -- minimum value [min]:                    -0.00370477 Jy/beam
(using all stokes and all channels)

HH211_I_zoom_4ch.image.pbcor
          -- maximum value [max]:                    0.0664254 Jy/beam
          -- minimum value [min]:                    -0.00370477 Jy/beam
(using stokes=I and all channels)

HH211_Q_zoom_4ch.image.pbcor
          -- maximum value [max]:                    0.00252461 Jy/beam
          -- minimum value [min]:                    -0.00225548 Jy/beam
(using stokes=Q and all channels)

HH211_U_zoom_4ch.image.pbcor
          -- maximum value [max]:                    0.0022621 Jy/beam
          -- minimum value [min]:                    -0.00238701 Jy/beam
(using stokes=U and all channels)

Now open HH211_IQU_zoom_4ch.image.pbcor and using default clip level=100%, we see the lower/upper bounds in colormap panel do not change with different stokes. This should be corrected.

I confirm that when a certain clip level is clicked (for example 95%), only the data at current stokes are used for calculations. This is verified by
1) open HH211_IQU_zoom_4ch.image.pbcor and switch to stokes=1 (Q) and set clip=95%.
2) reopen CARTA and open HH211_Q_zoom_4ch.image.pbcor and set clip=95%.
We should see identical lower/upper bounds in colormap. You may re-confirm my test. Nice work!

markccchiang:
There is a drawback as I mentioned in **Step2**. You have to click different percentage buttons twice then the cache will be refreshed (or trigged) for different Stoke. For example, open `HH211_IQU_zoom_4ch.image.pbcor` with clip=100% (default), switch to stokes=1(Q), click 95% and then click back to 100%. It will show the correct values. I will try to fix it (trigger mechanism) later.

kswang1029:
ok I see. Just checked following your steps.

carta/cpp/core/Algorithms/quantileAlgorithms.h
```diff
@@ -55,7 +55,8 @@ quantiles2pixels(
     std::vector < Scalar > allValues;
     view.forEach(
         [& allValues] ( const Scalar & val ) {
-            if ( ! std::isnan( val ) ) {
+            // check if the value from raw data is finite
+            if ( std::isfinite( val ) ) {
```
grimmerk: Need to confirm with @kswang1029 .

carta/cpp/core/Data/Image/Controller.cpp
```diff
@@ -361,7 +361,9 @@ std::vector<std::pair<int,double> > Controller::getIntensity( const std::vector<
 }
 
 std::vector<std::pair<int,double> > Controller::getIntensity( int frameLow, int frameHigh, const std::vector<double>& percentiles ) const{
-    std::vector<std::pair<int,double> > intensities = m_stack->_getIntensity( frameLow, frameHigh, percentiles );
+    int stokeFrame = getFrame(AxisInfo::KnownType::STOKES);
```
grimmerk, line 364: Is compatible with @appolloford 's change about `AxisInfo`?

carta/cpp/core/Data/Image/DataSource.cpp
```diff
@@ -453,42 +482,60 @@ std::vector<std::pair<int,double> > DataSource::_getIntensityCache( int frameLow
...
+                    // get the channel index corresponding to the above intensity value
                     intensities[i].first = allValues[locationIndex].first / divisor;
 
                     if ( frameLow >= 0 ){
                         intensities[i].first += frameLow;
                     }
 
                     // put calculated values in both the memory cache and the disk cache
-
-                    m_cachedPercentiles.put( frameLow, frameHigh, intensities[i].first, percentiles[i], intensities[i].second );
```
grimmerk, line 481: Need to add one more if to 
`leastRecentlyUsedCacheentry`::operator==

carta/cpp/core/Data/Image/DataSource.cpp
```diff
@@ -1015,22 +1153,31 @@ void DataSource::_updateClips( std::shared_ptr<Carta::Lib::NdArray::RawViewInter
 			maxClipInCache = m_diskCache->readEntry(maxClipKey.toUtf8(), maxClipVal);
 		}
 
-        if (minClipInCache && maxClipInCache) {
-            clips.clear();
-            clips.push_back(qb2d(minClipVal));
```
grimmerk: Do percentile and quantile both use these things?

carta/cpp/core/Data/Image/DataSource.h
```diff
@@ -304,6 +306,8 @@ class DataSource : public QObject {
      * @return the raw data or nullptr if there is none.
      */
     Carta::Lib::NdArray::RawViewInterface *  _getRawData( int frameLow, int frameHigh, int axisIndex ) const;
```
grimmerk: Note:
frameLow, and frameHigh are always -1, better to find a chance to investigate its purpose

carta/cpp/core/Algorithms/quantileAlgorithms.h
kswang1029: possible pixel values are 
1) regular float numbers
2) NaN

A frame may contain all pixels as NaN, or part of pixels as NaN.

A cube with all frames and all pixels as NaN is very rare.

carta/cpp/core/Data/Image/DataSource.cpp
grimmerk: Also, we need to figure out that 
1. does we really set diskcache part for each quantile?
2. why will the calculation of quantile for each channel not be executed for next time loop?

grimmerk: 3. also, it seems no memory cache here, need check ?

carta/cpp/core/Data/Image/DataSource.cpp
```diff
@@ -373,57 +373,85 @@ std::shared_ptr<Carta::Core::ImageRenderService::Service> DataSource::_getRender
...
                 qDebug() << "++++++++ intensity key is" << intensityKey.toUtf8();
 
                 if (intensityInCache) {
+
                     qDebug() << "++++++++ found location and intensity in disk cache";
                     intensities[i] = std::make_pair(qb2i(locationVal), qb2d(intensityVal));
                     foundCount++;
+
                     // put them in the memory cache
-                    m_cachedPercentiles.put( frameLow, frameHigh, intensities[i].first, percentiles[i], intensities[i].second );
```
grimmerk, line 411: Based on yesterday discussion, we still find a chance to study who will use this `intensities[i].first` and how.

grimmerk:
One more thing, @appolloford suggests to check if the added function of changing stokes works after permutating axes.

carta/cpp/core/Algorithms/quantileAlgorithms.h
markccchiang: If pixel values are +/-inf, the code would crash while loading pixel values in the function `std::nth_element()`. If pixel values do not have +/-inf for normal image files, I will change this line back to the original setting, i.e., `if ( ! std::isnan( val ) ) {...}`.

kswang1029: from my experience, i haven't seen pixel values as inf. It should have been set to NaN while creating the fits image.

kswang1029: will this inf check reduce performance?

carta/cpp/core/Data/Image/Controller.cpp
markccchiang: So far I do not find it will cause the problem.

carta/cpp/core/Data/Image/DataSource.cpp
markccchiang: Done!

carta/cpp/core/Data/Image/DataSource.cpp
markccchiang: In the original code, the author try save quantile results in `m_diskCache` first, but this part is not completely. So it leads to the failure of update (Max. and Min. pixel values) for different channel, that is why I cancel `m_diskCache`. There is another cache in this function called `m_quantileCache` which also stores quantile calculation results. So the memory cache still works if I cancel `m_diskCache`. The advantage of using `m_diskCache` is that maybe we can utilize the percentile results and don't need to calculate it again while switch to quantile mode (but it is vary rare from the current application). Now I have completed the  `m_diskCache` part and add this cache back to function.

carta/cpp/core/Data/Image/DataSource.h
markccchiang: `frameLow=-1` and `frameHigh=-1` mean we calculate the percentile for the whole channels. I guess the author has considered additional function that allows users to calculate the percentile for specific section in the whole channels. For example set `frameLow=0` and `frameHigh=2` that only calculate the percentile form channel 0 to channel 2.

carta/cpp/core/Algorithms/quantileAlgorithms.h
markccchiang: No. Just @grimmer0125 has a concern that it may filter out some information.

markccchiang:
Grimmer note: This below update is related to #87 

### Update: 

Show Quantile information on the Image Viewer if we turn on the Quantile mode, like the following figure:

carta/cpp/core/Data/Image/DataSource.cpp
grimmerk:
```
                    intensities[i].first = allValues[locationIndex].first / divisor;

                    if ( frameLow >= 0 ){
                        intensities[i].first += frameLow;
                    }
```

grimmerk: Colormap.cpp:
1st place to use this .first(which frame/channel), to prevent channcle from not postive. Seems not much useful.
```
    if ( intensities.size() == 2 && intensities[0].first >=0 && intensities[1].first >= 0 ){
        double minInt = intensities[0].second;
```

grimmerk: 2nd place, still need to check what its purpose is :

```
Colormap::_convertIntensity
    ...
    std::vector<double> valuesX(2);
    valuesX[0] = m_stateData.getValue<int>( INTENSITY_MIN_INDEX );
    valuesX[1] = m_stateData.getValue<int>( INTENSITY_MAX_INDEX );

->
             auto result = Globals::instance()-> pluginManager()
                            -> prepare <Carta::Lib::Hooks::ConversionSpectralHook>(image,
                                                  "", UnitsFrequency::UNIT_HZ, valuesX );
```

grimmerk (approved):

carta/cpp/core/Data/Image/DataSource.cpp
grimmerk: continue the above comment, there seem to be 2 or 3 places to use this hook or vector,

1. Colormap.cpp:
Colormap::_convertIntensity(
 ..
 convertedIntensity.first = data[0];

2. CurveData.cpp (profiler uses it):
                double roundedFreq = Util::roundToDigits( converted[0], significantDigits );

3. SpectralConversionPlugin.h (but it seems that no one uses this plugin)
                                        std::vector<double> inputValues = hook.paramsPtr->m_inputList;

Issue #156, opened 2017-05-23
"square" and "polynomial" scale function in colormap are identical


kswang1029:
this is perhaps by design or due to non-finished task in 
carta/cpp/CartaLib/PixelPipeline/CustomizablePixelPipeline.h

The power index of "polynomial" is hard-coded as 2, so it is identical to "square". 
Suggestion:
1) (takes time) make the power index of "polynomial" customizable by users 
or 
2) (easier way) change "polynomial" to "polynomial3" and hard-code the power index to 3. (we may include "polynomial4" as well and stop at 4.)

PR #158, opened 2017-05-23
generalize polynomial from second power to third and fourth power

kswang1029:
related to issue #156 

The scale functions "polynomial" and "square" are identical, therefore it is redundant.

I adopted an easier way by modifying the original polynomial (2nd power) to the 3rd and 4th power. The changes are isolated in three files. In the code, it is possible to have the freedom to let user to set the power index to the polynomial scale function but there will be extra work to add UI. Perhaps this freedom is not necessary to users (too much freedom). So I just generalized polynomial from the 2nd power to the 3rd and 4th power. Detailed re-scaling of colormap should be achieved by "gamma" value. Here we just provide a few handy options to users.

As an example, polynomial (N-th power) == linear + gamma=N.

### DIFF
--- a/carta/cpp/CartaLib/PixelPipeline/CustomizablePixelPipeline.h
+++ b/carta/cpp/CartaLib/PixelPipeline/CustomizablePixelPipeline.h
@@ -82,7 +82,8 @@ private:
 enum class ScaleType
 {
     Linear, /// x' = x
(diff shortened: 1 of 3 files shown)


grimmerk (approved):

grimmerk:
note:

The actual effect:
UI has two more options (third and fourth power) in colormap.

PR #160, opened 2017-06-03
unify font

kswang1029:
This PR tries to unify the font used for the UI. Instead of using the font from the OS (which may be system dependent), a local font file will be used.

The attached image demonstrates the result. Note that there is still difference in the appearance between mac (upper) and linux (lower) versions.
screen shot 2017-06-03 at 17 04 54


kswang1029:
update: the appearance inconsistency is likely due to x11 display (ssh). I tried docker and I see consistent results (both in desktop version and server version). It looks just like what I see in the Mac version.

grimmerk (approved):

Issue #97, opened 2017-04-28
now always use j2000 coordinate system as default system, regardless the system in file header


grimmerk:

low-sky:
What is the reasoning for this?  I'm thinking that Galactic frame images are best suited to being shown in Galactic coordinates by default.

kswang1029:
Current feature of carta is that regardless the coordinate system defined in header, what users see is always j2000. Coordinate transformation to j2000 will be performed for systems other than j2000. This is not what users expect to see. That's why this issue is opened and we shell fix it.

low-sky:
Ah, understood.  I didn't realize this was a Bug and not a Feature Request.

PR #163, opened 2017-06-07
set proper default coordinate system

appolloford:
This pull request mainly deals with the issue #97. There are roughly two parts:
1. Use the coordinate system type of direction coordinate(defined in casacore) from image files when loading images.
2. Provide a choice "Native" for users to change back to the default coordinate system of images after they change the coordinate system.
If there is no direction coordinate in the file, the coordinate system will be set to "Unknown" and cannot be changed.


carta/html5/common/skel/source/class/skel/widgets/Image/Grid/Settings/SettingsCanvasPage.js
```diff
@@ -196,6 +196,10 @@ qx.Class.define("skel.widgets.Image.Grid.Settings.SettingsCanvasPage", {
         _setCoordinateSystem : function( skyCS ){
             if ( typeof skyCS !== "undefined"){
                 this.m_coordSystem.setValue( skyCS );
+                this.m_coordSystem.setEnabled( true );
+                if ( skyCS === "Unknown" ){
```
grimmerk, line 200: Note: this is for PV case.

grimmerk (approved):

carta/html5/common/skel/source/class/skel/widgets/Image/Grid/Settings/SettingsCanvasPage.js
grimmerk: with "Unknown", if two files have different coordinates (not compatible), then do several times of switching, it is possible to produce an infinite loop then CARTA hang.

appolloford: The infinite loop didn't come from the incompatibility of two images. It's triggered by the undefined type of coordinate system when you changed back to, for example, a pv image from another image. CARTA cannot get a proper type to set. The "Unknown" type makes CARTA can know the fact that there is no coordinate system in an image. Therefore, it won't try to set the value of the coordinate system to a spaced string.

PR #166, opened 2017-06-08
New pre-defined layout: image composite including imageZoom and imageContext

grimmerk:


kswang1029:
tested with travisCI build 107, works well. (still imageZoom will start to show a zoomed image AFTER users zoom the image in the imageViewer. Default it is blank.) 

Suggest to merge this PR.

PR #177, opened 2017-06-30
Add timer to record the elapse time for opening an image file, extract percentile function to an individual file, and introduce two algorithms to speed up percentile calculations without image size limitation

markccchiang:
## Summary:

Since I did not expect that there are so many works and code changes done in this branch or PR when I started it, I just summarize modifications as following:

1. The elapse time for opening an image file including percentile calculations will be printed out and saved in the log file `carta.log` which is under user's home directory.

2. Extract the percentile calculation function to another individual file.

3. Add an iteration algorithm for computing percentile = 100% while opening image files. So we can reduce the time spending for opening image about 60%, compared to the original algorithm.

4. Add an algorithm to compute the approximation for percentile < 100%, if we set `"percentileApproximation" : "true"` in `config.json`. So we can speed up the percentile calculations and not be restricted by the size of image files. 

5. Users can also set the precision of approximate pixel value by `"percentApproxDividedNum" : 1000000` in `config.json`. The definition of precision for pixel value is `+/- (max-min)/percentApproxDividedNum`. The larger of `percentApproxDividedNum`, the more precise of pixel value with respect to the percentile we can get.

6. Solve the synchronize problem while change the `stoke index (I, Q, or U)` for `percentile calculation` on the `animation panel` and the image state dose not change at the same time.
 
7. Solve the synchronize problem while turn the `Clipping Per Frame` from true to false and the image state dose not change at the same time.


markccchiang:
Thanks for @confluence's comment! I have moved the checker `if (rawData == nullptr) {` right after when the rawData pointer is created.

grimmerk:
@markccchiang, @appolloford  , you can use CARTA_RUNTIME_CHECKS to judge if the code is built for release or not. If it is release, just not execute this testing/experimental block.

markccchiang:
## Update:

The sub-branch SHA: e796fc7cdb0964e6777820a0f548bfcc4dc3fc5a merges current `develop` and `confluence/intensity-location-cleanup` branch from PR #183. Also, according to @grimmer0125's suggestion, I have applied `CARTA_RUNTIME_CHECKS` to judge whether to print out the timer information in `carta.log`, if the code is built for release version or not.

markccchiang:
In 3297c5a64981ec1b57c07fc9372fc8e410a7670b, I add an iteration algorithm function `minMax2pixels()`, which is used for opening an image file or for computing the percentile=100%. Using this new function, we can increase the speed for opening image files as following (on the MacBook Pro):

`cube_x240_z2700_628MB_1.im`: 5445 ms --> 2023 ms (time spending: -62.8%)
`cube_x480_z900_831MB.im`: 6817 ms --> 2634 ms (time spending: -61.4%)
`cube_x960_z300_1106MB.im` 8983 ms --> 3460 ms (time spending: -61.5%)

markccchiang:
In d469d305283a0ccc49a742f1e435c6b162307cba, I add a function `percentile2pixels_approximation()` in `percentileAlgorithms.h`, which is the approximation for percentile (< 100%) calculation. It is used to deal with the memory problem for running CARTA with large image files on the single core computer. This function will be applied if the image file is larger than ~1.5G. `percentile2pixels_approximation()` ignores finding the possible location of approximate pixel value for percentile. It can be turned on by changing the variables `APPROXIMATION_GET_LOCATION = true` in `DataSource.cpp`. The performance (time spending) for CARTA with new algorithms for large image files are shown below (on the MacBook Pro):

Use original algorithm `percentile2pixels_I()` for percentile calculation (for small image files):

cube_x480_z900_831MB.im: 3379 ms (open image), ~5639 ms (for percentiles < 100%)
cube_x960_z300_1106MB.im: 3416 ms (open image), ~7818 ms (for percentiles < 100%)

Use new algorithm `peracentile2pixels_approximation()` for approximate percentile calculation (for large image files):

cube_x1920_z100_1474MB.im: 4465 ms (open image), ~9015 ms (for percentiles < 100%)
cube_x960_z2700_9959MB.im: 38301 ms (open image), ~80286 ms (for percentiles < 100%)
cube_x3840_z300_17695MB.im: 109717 ms (open image), ~203925 ms (for percentiles < 100%)
cube_x1920_z2700_39819MB.im: 348281 ms (open image), ~559162 ms (for percentiles < 100%)
cube_x3840_z900_53086MB.im: 440242 ms (open image), ~737479 ms (for percentiles < 100%)

To do: add optional items in UI. So users can decide whether to apply the approximate algorithm for percentile (< 100%) calculation, i.e. `APPROXIMATION_ELEMENT_LIMIT`. Also, allow users to decide the precision of approximation `APPROXIMATION_DIVIDED_NO`, which is the number to divide between minimum and maximum pixel values. The current setting is `APPROXIMATION_DIVIDED_NO = 1000000`.

markccchiang:
In 1ceda471d53cee71610320c47c7ef601d6216822, I set to apply the approximate percentile calculation for the file size larger than ~0.5G.

markccchiang:
In 58254e598d86dd754db267a9b32dcde2339b1a57, I solve the synchronize problem when turn the `Clipping Per Frame` from true to false and the image state dose not change at the same time. Also, I simplify the output of message `qDebug()` for code developers, so we can get important messages quickly.

markccchiang:
In 57a0392f046f94acaf6cfa16b9e9bb5157d62660, I add default settings whether to turn on the approximation algorithm for pixel values with respect to percentiles, and how precise of approximation for pixel value users want in the file `config.json`. If users set `"percentileApproximation" : "true"`, the approximation method for percentile calculation is turned on. If users set `"percentileApproximation" : "false"`, then the absolute precise algorithm for percentile `std::nth_element` is applied. Users can also set how precise of approximation method they want `"percentApproxDividedNum" : 1000000`. The definition of precision for pixel value is `+/- (max-min)/percentApproxDividedNum`. The larger of `percentApproxDividedNum`, the more precise of pixel value with respect to the percentile we can get.

Note that every time you change these settings in `config.json`, you have to delete the cache file in the directory `/tmp/pcache.sqlite`. Otherwise CARTA will reuse the old calculation results from previous configure settings. This part can be improved and let CARTA delete the old caches automatically in future.

confluence:
Instead of deleting the cache entries, we can implement the currently unused `priority` parameter of the cache functions: the priority should be based on the precision, and the highest priority cache entry should be retrieved. This should be straightforward to do in the SQLite plugin (add a new column and an `ORDER BY`), and you can leave it unimplemented in the LevelDB plugin for now.

markccchiang:
Thanks for @confluence's suggestion! I have modified the codes in the sub-branch 207d258956ee54ee1f54af164a4e6b5f5560b549 accordingly, which allows CARTA to save different percentile approximations in the cache and then select the best precision of pixel value for showing the image.

carta/cpp/core/Data/ViewManager.cpp
```diff
@@ -443,7 +489,31 @@ void ViewManager::_initCallbacks(){
         std::set<QString> keys = {Util::ID,DATA};
         std::map<QString,QString> dataValues = Carta::State::UtilState::parseParamMap( params, keys );
         bool fileLoaded = false;
+
+        if (CARTA_RUNTIME_CHECKS) {
+            // custmoize the style of print out messages
+            qInstallMessageHandler(myMessageOutput);
```
grimmerk: You can move `qInstallMessageHandler` to the constructor of ViewManager or coreMain.h, to be compatible with new CARTA which will multiple ViewManager, coreMain.h is better.

grimmerk (changes requested):

carta/cpp/core/Data/ViewManager.cpp
grimmerk: Also in new CARTA, writing the same log file should have thread safe protection. We may not need to correct to fit this requirement but we need this in new CARTA version.

grimmerk:
@markccchiang , does this pull request include `2nd point` of  https://github.com/CARTAvis/carta/issues/87 ?

markccchiang:
@grimmer0125 , No.

carta/cpp/core/Data/ViewManager.cpp
markccchiang: Thanks for @grimmer0125 's comments! It is already done.

confluence:
This isn't relevant right now, but when my unit conversion changes are merged, these algorithms should also take unit conversions into account.

It should be straightforward to adapt the loop from the exact intensity calculation which copies data to the loop in `percentile2pixels_approximation`.

An adjustment also needs to be made in `minMax2pixels` -- it wouldn't be sufficient to perform the conversion on the values at the end, because a frame-dependent conversion may change which image values are the minimum and maximum. It should be possible to avoid repeated calculations by doing something similar to what the intensity -> percentage function's comparison loop does.

A converter provides a constant scaling factor and possibly also a frame-dependent conversion function. For each frame, you apply the inverse of the frame-dependent conversion for the *current* frame to the previous min and max. Then all the raw values for that frame can be compared to the previous min and max. If a new min or max is found, store the new value with the frame-dependent conversion applied. The constant scaling factor can be applied once at the end.

All references to calculating the location can also be removed -- so only `percentile2pixels` should be kept; the function which is currently called `percentile2pixels_precise` should be deleted; `minMax2pixels` should return only intensities.

I'm not sure in which order these two branches will be merged, and where it makes the most sense to make these changes.

An unrelated suggestion for `minMax2pixels`: instead of checking afterwards whether any of the requested quantiles are 0 or 1, eliminate the `quant` parameter entirely, and return a pair or 2-element vector of intensities. Then explicitly assign the output as the intensities for the 0 and 1 quantiles in the calling code.

I would also suggest that instead of passing the min and max into `percentile2pixels_approximation` you calculate them inside the function, use those values for the 0 and 1 quantiles if they were requested, and if there are no other quantiles to calculate, return immediately without creating the histogram.

In the calling code in DataSource in the short term you can hard-code setting a higher priority for the cached intensity values for 0 and 1 because you know that they will have been calculated exactly no matter what -- but it would probably be more correct in the long term to return certainties/priorities as well as values from the algorithms, so that the calling code can set a priority without knowing the algorithm implementation, and perhaps without even knowing what algorithm was used (because it may have been selected elsewhere on the basis of the file size, configuration, etc.).  I will probably look into adding that when I implement my algorithm.

confluence:
FYI, I intend to write helper functions for iterating over the image and applying a function to each value, so the different algorithms won't have to duplicate the iteration code (which would be very annoying). So maybe don't worry about this for now, and use the helper functions from my new branch once it's complete.

markccchiang:
Thanks for @confluence's comments and suggestions! I am reading your codes in PR #200 and thinking how to merge these two branches.

confluence:
I just pushed a commit to that branch which cleans some stuff up and hopefully makes these elements easier to reuse.

grimmerk:
@markccchiang , have you updated the information about the optional setting of config.json to @kswang1029 ? It may be included in the manual. 

Also, I suggest you either draw a flow chart or even write unit tests for this histogram approximation method. You can use @confluence 's branch as a reference to write unites.

confluence:
You can use my branch as a reference for using the Qt test framework, although the documentation [here](https://doc.qt.io/qt-5/qtest-overview.html) and [here](https://doc.qt.io/qt-5/qttestlib-tutorial1-example.html) is pretty good. The tests I wrote were for a plugin, which required some additional integration steps that you don't need to worry about if you're writing tests for a part of the core application. The existing core tests use the Catch framework -- it may be simpler for now just to use that, although I think we should eventually migrate all the tests to the same framework.

markccchiang:
Thanks for good suggestions from @grimmer0125 and @confluence! I will work on these parts.

PR #181, opened 2017-07-07
Added the function of opening file by dbclick

appolloford:


grimmerk (approved):

kswang1029:
double-click to open an image will not be a feature in the new version. There should be buttons for users to select which mode to render such as raster, contour, vector, marker, etc. please consider to revert the commit if this part of code will be reused in the new version.

PR #183, opened 2017-07-14
Refactor functions for retrieving intensities corresponding to percentiles

confluence:
This PR introduces the following changes:

1. Cherry-picked commit 7990f0d, which moves the percentile algorithm code out of `DataSource` and into a new file in `Algorithms`

2. Removed `quantileAlgorithms` files, which define one function which is a duplicate of a function in `percentileAlgorithms` and a function which is unused.

3. Removed unnecessary internal redirection from `_getIntensity` to `_getIntensityCache` (how the cache is used is an implementation detail).

4. After discussion in Slack, removed the old LRU cache implemented in `LeastRecentlyUsedCache`, as well as the equivalent caching functionality used in `_getQuantileIntensityCache`, which now uses the same disk cache as `_getIntensity`.

5. Made algorithms return maps with percentiles as keys rather than vectors, to make it simpler to calculate a subset of the requested percentiles and reconstruct the original ordering. Fixed bug in cherrypicked code.

6. Made various small changes to make the two algorithms more consistent.

7. Renamed every `getIntensity` function to `getIntensityAndLocation`, and added new `getIntensity` functions which return only intensity values.

8. Corrected all references to these functions, so that only the Colormap code, which requires the location, calls `getIntensityAndLocation`.

Still to be done:

1. Unify `_getIntensity` and `_getQuantileIntensityCache`, which duplicate a lot of functionality.

2. Determine how the `Colormap` code which currently requests a location value is intended to function, whether it is correct, and whether this calculation could be eliminated entirely.


carta/cpp/core/Algorithms/percentileAlgorithms.h
```diff
@@ -0,0 +1,190 @@
...
+    if ( allValues.size() == 0 ) {
+        qFatal( "The size of raw data is zero !!" );
+    }
+
+    std::map < double, Scalar > result;
+
+    // for every input percentile, do quickselect and store the result
+
+    for ( double q : quant ) {
+        size_t x1 = Carta::Lib::clamp<size_t>( allValues.size() * q, 0, allValues.size()-1);
```
markccchiang: Propose to replace the `line 79` as following:
```
        // "size_t" can not be -1 in "Carta::Lib::clamp" function, which can not be used in "nth_element" searching for the minimum value
        //size_t x1 = Carta::Lib::clamp<size_t>( (allValues.size()*q)-1, 0, allValues.size()-1);
        int x1 = Carta::Lib::clamp<int>((allValues.size()*q)-1, 0, allValues.size()-1);
```

carta/cpp/core/Algorithms/percentileAlgorithms.h
```diff
@@ -0,0 +1,190 @@
...
+        timer2.start();
+
+        //std::nth_element( allValues.begin(), allValues.begin()+x1, allValues.end(), compareIntensityTuples );
+        std::nth_element( allValues.begin(), allValues.begin()+locationIndex, allValues.end(), compareIntensityTuples );
+
+        // end of timer for getting the nth_element from the raw data
+        qCritical() << "<> Time to get the nth_element from the raw data:" << timer2.elapsed() << "milliseconds";
+
+        //result.push_back(std::make_pair(allValues[x1].first/divisor, allValues[x1].second));
+        result[q] = std::make_pair(allValues[locationIndex].first/divisor, allValues[locationIndex].second);
```
markccchiang: Propose to replace the `line 160 to 181` as following:
```
        // "size_t" can not be -1 in "Carta::Lib::clamp" function, which can not be used in "nth_element" searching for the minimum value
        //size_t x1 = Carta::Lib::clamp<size_t>( (allValues.size()*q)-1, 0, allValues.size()-1);
        int x1 = Carta::Lib::clamp<int>((allValues.size()*q)-1, 0, allValues.size()-1);
        CARTA_ASSERT( 0 <= x1 && x1 < allValues.size() );

        // start timer for getting the nth_element from raw data
        QElapsedTimer timer2;
        timer2.start();

        std::nth_element( allValues.begin(), allValues.begin()+x1, allValues.end(), compareIntensityTuples );

        // end of timer for getting the nth_element from the raw data
        qCritical() << "<> Time to get the nth_element from the raw data:" << timer2.elapsed() << "milliseconds";

        result[q] = std::make_pair(allValues[x1].first/divisor, allValues[x1].second);
```

markccchiang (commented):
Thank you for your modifications! It improves the cache functionality a lot and makes the code much more clear. I would like to propose additional modifications in `percentileAlgorithms.h` which makes `percentile2pixels()` and `percentile2pixels_I()` functions coherent for searching the minimum and maximum pixel values. The detail of my proposals are shown in following code comments. Please check if they are correct.

carta/cpp/core/Algorithms/percentileAlgorithms.h
confluence: Surely if the minimum value is needed, the rank should be 0, not -1? I'm not sure if `allValues.begin() - 1` is even a valid parameter to `nth_element`. That's before the start of the vector. So the type change shouldn't be necessary.

I agree that in general the rank of the nth element should be calculated as `(size * q) - 1`, not `size * q` (although for a sufficiently large dataset off-by-one differences in the rank are unlikely to make much of a difference).

`percentile2pixels_I()` appears to do this more correctly, by calculating `(size * q) - 1` and clipping the bottom value to 0, whereas `percentile2pixels` calculates `size * q` and clips the top value down to `size - 1`.

I do propose that we pick one or the other for consistency; it seems sensible to pick `(size * q) - 1`.

confluence: Oh, wait, I think I misunderstood what you meant. I see that the value in your proposed code is still clipped down to 0 -- is the problem that a value of -1 cannot be passed as a parameter to `clamp<size_t>` because it expects `size_t` values and a `size_t` value cannot be -1?

I am still concerned about converting this value to int for the sake of one edge case, since we would lose half the available range (and it would make the comparisons in the CARTA_ASSERT more difficult). I would prefer to handle the edge case explicitly. I will implement my alternative proposal and commit it to the PR.

confluence: Actually, no special handling is necessary. The cleanest way to do this is:

    size_t x1 = Carta::Lib::clamp<size_t>(allValues.size() * q , 1, allValues.size()) - 1;

Now all the variables are >= 0 in all cases.

confluence: I have now committed my proposed solution.

markccchiang: Thanks a lot for your solution! I am good with it.

markccchiang (approved):

kswang1029:
any suggestion for testing it from user point of view (i.e. me)?

markccchiang:
From user's point of view, it should be almost the same as before. I will try to merge the other modifications  from PR #177.

kswang1029:
ok thanks. I will do same test as before, using travis build "#145"

markccchiang:
Hi @kswang1029, I have merged this branch to PR #177, you can try it. I forgot to mention that @confluence's modifications prevent the potential cache errors and make cache accesses more efficient. It also corrects a little difference between percentile and quantile modes.

grimmerk:
@confluence , since this pull request does not have an issue. So I just change the assignee of this PR to @kswang1029 to test. After he verifies, we can merge this pull request. 

If you are using https://app.zenhub.com/ or its chrome/firefox plugin, you can see the pipeline becomes `ReadyForQA`.

We can consider to integrate Zenhub with Slack, https://www.zenhub.com/blog/zenhub-slack/

confluence:
Great! Thanks!

kswang1029:
verified. Thanks for the hard work!

PR #182, opened 2017-07-12
Revert "Menu bar button icon fix (minor UI adjustment)"

grimmerk:
Reverts CARTAvis/carta#175

After testing on 
Mac, Qt 5.3 & Qt 5.7
Linux, Qt 5.6
There is are two side effects, 
1. right click of mouse becomes not working. 
2. Clicking on Data/Session on menubar will not open their menu.

Here is the screenshot of the log, 

default

So just revert it first. @JamieDass, I'm sorry that I test it late, I was waiting for fixes of travis-ci.

PR #184, opened 2017-07-14
fix for menubar button items (Version 2)

JamieDass:
This is a second attempt at fixing a minor UI feature that was attempted in:
https://github.com/CARTAvis/carta/pull/175

Unfortunately it was slightly buggy and reverted in https://github.com/CARTAvis/carta/pull/182.

I believe this new commit addresses the bug.


kswang1029:
verified in docker-ubuntu. Now the pan reset/zoom reset buttons are fixed as shown in the attached image.
screen shot 2017-07-18 at 1 37 29 pm

PR #185, opened 2017-07-17
Fix for broken plugins in Travis-CI builds (missing libgsl.19.dylib)

ajm-ska:
Homebrew updated its gsl version to 2.4. This version no longer provides libgsl.19.dylib which is currently used by some other cached libraries in our Travis-CI setup. 
The result is that some plugins (e.g. profiler and image statistics) will fail to work in the final package.

A temporary workaround is to build the older gsl 2.3 from its source code instead.

### DIFF
--- a/carta/scripts/ci_mac_common.sh
+++ b/carta/scripts/ci_mac_common.sh
@@ -85,8 +85,15 @@ function installqt() {
 
 function installgsl() {
 su $SUDO_USER <<EOF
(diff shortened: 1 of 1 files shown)


grimmerk:
@ajm-asiaa , since this pull request includes some experimental commits, and this is not good to merging to `develop`, so I overwrite the history and force push.

PR #186, opened 2017-07-23
Add CartaLib directory to libcore RPATH / RUNPATH to fix linker issue on Ubuntu 17.04

confluence:
Since version 2.28, `ld` (on Debian and possibly other Linux systems) creates a RUNPATH instead of an RPATH (in response to [this issue](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=835859)). Because RPATH and RUNPATH are [resolved differently](https://blog.qt.io/blog/2011/10/28/rpath-and-runpath/), this in combination with the way that `libcore.so` is built makes it impossible to run the CARTA executable from the build directory on Ubuntu 17.04 (without making more additions to LD_LIBRARY_PATH at runtime) because the executable cannot find `libCartaLib.so`.

This commit fixes the link from `libcore.so` to `libCartaLib.so`, which fixes the issue on Ubuntu 17.04, and does not appear to impact the build on e.g. Ubuntu 16.04, which uses an older `ld` which creates RPATHs rather than RUNPATHs.

I'm not sure if this change has any impact on the final packaging process, and it has not been tested on OSX.

### DIFF
--- a/carta/cpp/core/core.pro
+++ b/carta/cpp/core/core.pro
@@ -342,6 +342,8 @@ INCLUDEPATH += $$absolute_path(../../../ThirdParty/rapidjson/include)
 #INCLUDEPATH += ../../../ThirdParty/qwt/include
 #LIBS += -L../../../ThirdParty/qwt/lib -lqwt
 
+QMAKE_LFLAGS += '-Wl,-rpath,\'\$$ORIGIN/../CartaLib\''
+
 QWT_ROOT = $$absolute_path("../../../ThirdParty/qwt")
 INCLUDEPATH += $$QWT_ROOT/include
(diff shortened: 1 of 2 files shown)


grimmerk:
Here is the checklist. 
- [x] build on Mac. Checked. 
- [x] packaging on Mac
- [x] build on CentOS
- [x] packaging on CentOS 
@ajm-asiaa , can you try the final one? Just make sure this will not affect the packaging (perl) script running.

grimmerk (approved):

ajm-ska:
I confirm that it works with packaging on CentOS.

grimmerk:
@ajm-asiaa , thank you.

Issue #105, opened 2017-05-02
lag cursor info when region exists in the server version with docker


kswang1029:
this ONLY happens in the server version with docker. I also tested both desktop and server version in CentOS7, but I don't see it.

kswang1029:
Solved by #187

PR #187, opened 2017-07-26
Remove hover behaviour for regions

JamieDass:
This change removes the hover behaviour for regions. Previously, hovering over a region would highlight its control points etc but at a performance cost. This behaviour does not really seem necessary.
This change resolves the cursor lag reported in [issue #105](https://github.com/CARTAvis/carta/issues/105)

### DIFF
--- a/carta/cpp/core/Data/Region/Region.cpp
+++ b/carta/cpp/core/Data/Region/Region.cpp
@@ -15,7 +15,6 @@ namespace Carta {
 namespace Data {
 
 const QString Region::ACTIVE = "active";
-const QString Region::HOVERED = "hovered";
 const QString Region::COLOR = "color";
 const QString Region::REGION_TYPE = "regionType";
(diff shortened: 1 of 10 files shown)


kswang1029:
is this remove action the only solution to fix the performance issue? Actually in casaViewer, when multiple regions present, a region will be highlighted when hovered.

JamieDass:
As far as I can see, with the way that Regions are created and managed in Carta right now, removing the behaviour is the only way to solve this performance lag. I don't think we lose much by removing it.

In the future it might make sense to rewrite the underlying Region code and it could be reintroduced if that happens but that's beyond my scope.

low-sky:
I agree that this doesn't make much sense to polish given the impending changes to UI and architecture.

kswang1029:
OK. Verified. Propose to merge to dev. Thanks for fixing this.

PR #193, opened 2017-08-07
Preserve the directorty path accessed at the last time opening filetree

appolloford:
This PR preserves the directory path of filetree when opening filebrowser and uses it to replace the default path. @kswang1029 thinks that this is more convenient.


kswang1029:
verified. good job.

grimmerk:
This current change may result in a side effect:

In CARTA server version, B user will see the last path which is made by A user.

grimmerk:
After studying the code, it seems that the quick solution is to store some information in Browser Clients. Either cookie or `HTML5 Local Storage`.

grimmerk:
@appolloford Can you change it to non-static? Non-static will work in current CARTA and new CARTA.

PR #195, opened 2017-08-22
making point region highlighted when selected and other cosmetics tweaks

kswang1029:
Based on @JamieDass 's region cosmetics improvement of point region, this PR tries to make point region highlighted when being selected, like other shapes of region. Originally point region appears to be selected even if it is not. Attached image demonstrates the results. Colors are changed to high contrast and high saturation ones for better visibility. Line width is changed to thinner one. Control point is changed to smaller size. The size of the point region is also changed to smaller one. 

screen shot 2017-08-22 at 9 40 41 pm

TODO: ideally the size of the control point and "point" region should be independent of image pixel size and image zoom level. Now they have fixed sizes. So when either image size is small (<100 px) or at high zoom level, the appearance of control point and "point" region will be too large. 

Note: commit 970d279 is only for making travis CI build successful.

### DIFF
--- a/carta/cpp/CartaLib/Regions/Point.cpp
+++ b/carta/cpp/CartaLib/Regions/Point.cpp
@@ -37,7 +37,7 @@ bool Point::isPointInsideUnion( const RegionPointV & pts ) const {
 
 QRectF Point::outlineBox() const {
 	QRectF box;
-	const int SMALL_SIZE = 11;
+	const int SMALL_SIZE = 10;
 	double topLeftX = m_point.x() - SMALL_SIZE / 2;
 	double topLeftY = m_point.y() - SMALL_SIZE / 2;
 	box.setX( topLeftX );
--- a/carta/cpp/CartaLib/Regions/Point.cpp
+++ b/carta/cpp/CartaLib/Regions/Point.cpp
@@ -37,7 +37,7 @@ bool Point::isPointInsideUnion( const RegionPointV & pts ) const {
 
 QRectF Point::outlineBox() const {
 	QRectF box;
-	const int SMALL_SIZE = 30;
+	const int SMALL_SIZE = 11;
 	double topLeftX = m_point.x() - SMALL_SIZE / 2;
 	double topLeftY = m_point.y() - SMALL_SIZE / 2;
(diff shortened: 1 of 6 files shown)


JamieDass:
Hi @kswang1029,

This looks good. I originally wanted to leave the point outline visible just because it's hard to see otherwise but it is confusing that this is inconsistent with selection/deselection as it is for other shapes so the change makes sense.
These colours are still hard to see for me at least in the bottom left panel where you show the point regions, but I think no matter what colour we choose it will be difficult in certain situations. This is why I added the option to change the region's colour.

As for the TODO you mentioned about having a scale-independent point region: this is something that I've thought about quite a bit and unfortunately I don't see a good way to do it. There are only 2 options as far as I can see:

1.  make it a "zero"-width point — much like the lines depicting regions which are scale-independent. There is a parameter in QTPainter for this.
2. Scale the shape with zoom. Redrawing is quite costly in terms of performance but I haven't tested it.

kswang1029:
@JamieDass Sure, there is no perfect color for regions since colormap can be any. Here I just made the colors to have high saturation and contrast at least. This little change should be better than the original ones. If the colors are not good for certain colormap, then your implementation of changing the colors of region comes in and it should help a lot. 

I will try to see if I can make a scale-independent point and see how it appears overall. Thanks for the suggestions.

kswang1029:
@JamieDass I have changed the linewidth of point region scale-independent. Now the line behaves just like other region shapes. Please refer to the attached image. It's dimension is 512 x 1024 px. 

screen shot 2017-08-23 at 3 37 17 pm

kswang1029:
How about this approach: I changed control point from filled rectangle to open rectangle with its lineWidth independent of scale. Open rectangle helps users to see clearly the pixels near the control point. Therefore, at high zoom level users can still see the image, rather than being blocked by the control point.
screen shot 2017-08-23 at 8 26 49 pm

JamieDass:
That's a nice workaround!
It's a shame we can't make the size of the points smaller in a nice way but I think this is the best solution.
After the new architecture is implemented, we could maybe look at scaling the points with the zoom/image size and see how much performance decreases.

PR #199, opened 2017-08-30
[Mark/contour] Add more contour types and try to improve contour performances

markccchiang:
This PR is used to test the speed of calculating contour with different contour types. The usage of choosing and plotting contours with different types are shown below. By default, we use the contour type `w/o line combine optimization` to calculate contours, which is the fastest one. The original contour type in CARTA is `w/ line combine optimization`, which spends at least twice of time to plot contours compared to `w/o line combine optimization`. Users can check the difference between these two types. I also add the other contour types `w/ Gaussian blur 3x3`, `w/ Gaussian blur 5x5`, `w/ Box blur 3x3` and `w/ Box blur 5x5` in `Contour Type/Sets:` item. Time spendings for the other contour types are roughly same with `w/o line combine optimization`.

img1
img2
img3

figure_2
figure_1

Users can add different contour sets with different ranges and colors by following steps: 

img4
img5
img6

**To do: extract and create a new `Contour Type` item from `Contour Type/Sets:` on the UI panel.**


markccchiang:
### Update
1. Create a new item `Smoothness` on the contour UI.
2. Set the default Smoothness of contour as `Gaussian blur 3x3`.
3. Set the default color of contours as `light blue`.





**Todo: if possible, allow users to set different contour Smoothnesses for their specific contour levels (in different contour sets)**

kswang1029:
Thanks. Let's lower the priority of the TODO for now.

markccchiang:
### Todo list (in priority order):
1. Add a cache to save contour points so that contour movie in `fps` can be increased further after the second run of cycle.
2. Synchronize the default of contour min/max with percentile min/max.
3. Improve and optimize functions in `ContourConrec.cpp`.

carta/cpp/CartaLib/Algorithms/ContourConrec.cpp
```diff
@@ -284,6 +281,509 @@ conrecFaster(
...
+                          }
+                          );
+            CARTA_ASSERT( i == nCols );
+        };
+        updateRows(); // update the row number: n + 1
+        updateRows(); // update the row number: n + 2
+        updateRows(); // update the row number: n + 3
+
+        // to keep the data accessor easy, we use this lambda, and hope the compiler
+        // optimizes it into an inline expression... :)
```
appolloford, line 337: 1. I think that it is better if we can multiple the gaussian weight when we do the updateRow. Another way I think is using updateRow to get data with size 4*col, and use a function multiplying gaussian weight to transform the data to size 2*(col-2). Then the following codes can be reused.
2. If the original author hopes the lambda become an inline function, maybe we can directly claim it to be an inline function.

carta/cpp/CartaLib/Algorithms/ContourConrec.cpp
```diff
@@ -284,6 +281,509 @@ conrecFaster(
...
+
+        Carta::Lib::Algorithms::ContourConrec::Result result;
+        if ( nc < 1 ) {
+            return result;
+        }
+        result.resize( nc );
+
+    #define xsect( p1, p2 ) ( h[p2] * xh[p1] - h[p1] * xh[p2] ) / ( h[p2] - h[p1] )
+    #define ysect( p1, p2 ) ( h[p2] * yh[p1] - h[p1] * yh[p2] ) / ( h[p2] - h[p1] )
+
```
appolloford, line 353: Maybe the following C-type claim can be changed to C++-type claim.

carta/cpp/core/DefaultContourGeneratorService.cpp
```diff
@@ -48,10 +55,20 @@ DefaultContourGeneratorService::start( Lib::IContourGeneratorService::JobId jobI
 void
 DefaultContourGeneratorService::timerCB()
 {
+    // add timer
```
appolloford, line 58: Maybe we can add CARTA_RUNTIME_CHECK

markccchiang: Done.

carta/cpp/CartaLib/Algorithms/ContourConrec.cpp
markccchiang: Thank you for the suggestions! Since this function is highly correlated, I have no good ways to trim it so far. If we multiply weights in`updateRow()`, the raw data is changed and can not be reused for the next iteration. If we modify the lambda function `acc(i,j)` as an inline function, it seems the code would become more complex and the calculation speed is not improved significantly. So I tend to take this as in the TODO list.

carta/cpp/CartaLib/Algorithms/ContourConrec.cpp
markccchiang: Done.

Issue #89, opened 2017-04-28
no (intensity) unit conversion


grimmerk:
related doc: https://docs.google.com/document/d/1VZ64s2V47mJzXX4YBBTDN6jORu9XQKc6WFy3QeCdviU/edit
*Need to consider multi-beam during conversion.* 

if the unit conversion is not frame-property dependent, such as Jy/arcsec^2 to MJy/sr, the bound values (such global min/max or the min/max per frame or clip level at non-100% levels) can be directly converted to new unit which is also not frame-property dependent. 

As long as the unit conversion dependents on frame properties, such as frequency and/or beam size (which may also change from frame to frame), conversion of all pixels needs to be done first before running percentile/quantile.

kswang1029:
@confluence let's continue the unit related discussions here.

confluence:
Good idea.

I am working on this in a branch now. My plan is as follows:

1. Create unit tests for the `ConversionIntensity` and `ConversionSpectral` plugins

2. Refactor and clean up the plugin code

3. Change the plugin API so that instead of accepting values, units and other data and returning converted values the plugin accepts units (and other data) and precalculates and returns a simple function / lambda which can be called repeatedly on values to be converted (possibly all the values in the image). Each conversion function which needs to be called on every frame + value pair (because the conversion depends on the frame) should ideally also be split into the portion which actually depends on the frame and any constant multiplier which can be applied once afterwards. (I.e. we should never be multiplying by a constant scaling factor inside the loop iterating over the image.)

4. Remove the current incorrect percentile calculation with location

5. Add unit awareness to the percentile calculation without location: when the percentiles are calculated for a unit which depends on the frame value, the conversion must be performed on all the values in the image before the calculation. This should be done with the function(s) returned from the plugin: a frame-dependent conversion to be called on each value individually (which may be nothing, if the conversion is not frame-dependent) and a constant multiplier which can be applied once to the final value. The percentile calculation would be performed only on intensity values; the frame value should only be used in a possible pre-conversion and then discarded. This conversion can be integrated both into the exact algorithm (where the finite values are copied to a new vector), and the approximate algorithm I plan to add (which also iterates over the image).

6. Add unit awareness to the cache: the percentiles must be recalculated and cached for every unit which cannot be derived from an existing cached unit by scaling. We should decide whether to cache only one unit per category and always scale from cached base units or whether to cache all final units.

confluence:
[...]

PR #200, opened 2017-09-04
Confluence/unit conversion cleanup

confluence:
This PR partially fixes the unit conversion-related issues described in #89.

Notable changes:

* The function for calculating an intensity value with a location for a percentile has been removed.

* Functions for calculating intensities from percentages and vice versa take an optional intensity converter parameter and an optional Hertz values parameter, and can handle frame-dependent unit conversions (currently this is only implemented for conversions to and from Kelvin). Approximate algorithms should use the same parameters.

* The intensity cache now caches values separately for each base frame-dependent transformation (currently there are three options: no transformation, and the two transformations associated with conversions to and from Kelvin). Caches should probably be cleared when this change is merged, to remove old values which are no longer usable (although there should be no key name clashes if they are kept).

* The intensity conversion plugin has been refactored and its API has changed. It now returns a conversion function (wrapped in an object with some other information). 

Some known outstanding issues:

* Some unit conversions are unsupported, e.g. to/from Jy/pixel

* Frame-dependent beam information is not supported

* Unit conversion is not supported in the pixel pipeline, which affects frame-dependent conversions; it's currently not possible to e.g. convert a Jy/Beam image to Kelvin, calculate the quantiles in Kelvin, and correctly apply them to the image display. Intensities are always recalculated in the image's original units, so changing units has no effect on how the image is rendered.

* The spectral conversion plugin should also be refactored to have an API similar to the intensity conversion plugin (i.e. accept units and return a function, not accept units and original values and return converted values) -- this would allow some of the code to be cleaner.


confluence:
When I've finished working on my approximate quantile algorithm I will rebase this branch on the latest development branch and make any updates that are required to integrate it with @markccchiang's most recent merged changes. Please disregard this PR until then.

confluence:
I believe that this is now up-to-date with current develop and can be merged. @markccchiang, do you have any comments about the merge with your changes? Mostly I removed all the location-related code, and replaced the additional function in DataSource for getting the min and max intensities before the approximate algorithm with a recursive call to the main intensity function (which already has an option for using the min/max algorithm). I have some other thoughts about simplifying the code, but I didn't want to overcomplicate this PR with too many changes that are not directly relevant. I will suggest these in smaller PRs later.

Would it be possible for you to rebase your branch with the test on these changes and check that the approximation code is still working as expected? I did some manual tests with and without unit conversion with approximation turned on and everything seemed OK, but you are more familiar with this code.

markccchiang:
@confluence Thanks a lot for your works. This is a significant improvement for the CARTA. I am checking your codes and hope it can be merged in before the other PRs related to the quantile function (as you have suggested).

carta/cpp/core/Data/Colormap/Colormap.cpp
```diff
@@ -232,77 +239,92 @@ QString Colormap::_commandSetColorMap( const QString& params ){
...
+
     if ( controller ){
+        QString fromUnit = controller->getPixelUnits();
+        
+        // This allows clips and intensity bounds to be changed without a converter if the image has unknown units
+        if (fromUnit == "N/A" || toUnit == "N/A") {
+            qWarning() << "Image units are unknown. Returning null intensity converter.";
+            return intensity_converter;
+        }
+
```
markccchiang: It seems the conversion of the unit "Jy/pixel" is not available for now. If the unit of an image file is "Jy/pixel", the code will throw an exception and the percentile calculation cannot be proceeded by "Clipping" button. To allow the "Clipping" button works, can we treat the unit "Jy/pixel" as the same with "N/A" for an image file like adding the following codes?
```
        // This allows clips and intensity bounds to be changed without a converter if the conversion of image unit is not available.
        if (fromUnit == "Jy/pixel" || toUnit == "Jy/pixel") {
            qWarning() << "No function available to convert the unit Jy/pixel. Returning null intensity converter.";
            return intensity_converter;
        }
```

confluence: Instead of hardcoding exceptions for unavailable units (of which there are many), I would prefer to return a null converter instead of raising an exception whenever a converter cannot be found -- if what we want is for clipping to work always even if the file units are not recognised or are missing data necessary for a conversion.

So I could also return a null converter instead of raising an exception if the conversion requires hertz values but none are available (because the image lacks a spectral axis, or because the spectral units were not recognised).

We would just need to watch out for possible side effects, like unit changes being shown in the UI when the unit conversion did not succeed (right now we handle this by not showing unavailable units in the menu, but IIRC it's not easy to check whether the hertz values are going to be available at the point where that menu is populated, which is why there's a separate check for it). I don't know if it's possible to prevent the unit value from changing in the UI if the units were not changed -- I have looked at this before but I don't know the UI code well enough to fix it. If we are changing the UI anyway this may not be an issue in the long term.

I will commit this change -- please let me know if you foresee any issues.

carta/cpp/plugins/ConversionIntensity/ConverterIntensity.cpp
```diff
@@ -1,403 +1,176 @@
 #include "ConverterIntensity.h"
-#include <plugins/ConversionSpectral/Converter.h>
 #include <math.h>
```
markccchiang, line 2: Another thing is that could we add an additional code line `#include <cmath>` here? It happens to me when I am compiling the code on the Mac OS, I got an error message that no member named `round` in namespace `std` for the code line 161: `exponent += std::round(log10(rx.cap(1).toInt()));`. Adding `#include <cmath>` can resolve this compiling error.

confluence: Header added.

markccchiang (approved):
Thanks for your prompt response! There are no further questions for this PR. We can merge it before the other PRs. Next step I will refactor my percentile function and move it to the plugin module. Please remind me if there are suggestions or ideas.

confluence:
Thanks! My suggestions for the plugin refactoring are:

1. Have each algorithm (or plugin; see below) return the error as well as the value. Then the core code doesn't have to know anything about the implementation of the plugins. I was initially thinking that there should be one error per value, in case they are different -- for example, your approximation uses exact min and max values. But those are calculated a level up in the calling code anyway so that they can go through the cache, so perhaps that is not necessary right now (we can change it later if we need it). I think the cleanest way to do this is to include it in a custom class returned from the plugin.

2. I haven't yet done a proper comparison of our two approximate algorithms and how they perform on different images -- if it turns out that the two algorithms perform better or worse on images of a particular size or with some other property, we may want to enable both of them and decide which one to use at runtime. My idea for handling this was to have each plugin examine the properties of the image and return a function (like the refactored intensity conversion plugin), with some kind of score, so that the calling code can compare the scores and use the "best" plugin for that image.

3. This information should be encapsulated in a class (which inherits from an interface defined in the core, so that the function parameters are the same for each plugin -- this may make it a good idea to put the exact algorithm, which should stay in the core, inside a class inheriting from the same interface). This object can also include the error for the algorithm.

I was planning to do this for my approximate algorithm -- if you like, I can do the implementation for both.

markccchiang:
Your suggestion sounds good! So far I have no idea how to score or evaluate an image file and decide which algorithm plugin is the best for it. If we can get some useful information from the header of an image file that is good. Otherwise, this evaluation may cost additional time. Allowing users to decide the algorithm or let the code decides could be an option in `config.json` setting. I think you are more familiar with the plugin interface. If it is straightforward for you to apply my algorithm in the new plugin feature, you can do me a favor. Thanks a lot!

confluence:
I'm assuming that the evaluation will be something simple and inexpensive, like checking the image dimensions or other header properties. Of course the user can disable a plugin entirely to make sure that only one plugin will ever be used. We can also select the plugin only once per image and cache the decision (I don't foresee the need for the algorithm to vary depending on specific percentiles).

I'm happy to work on this; that was what I was planning to do next. Once this PR is merged, could you also update and merge your algorithm test? I would like to incorporate my algorithm into that too, so that I can compare the performance of the two approximations on different image files. I will also add your algorithm to my unit test (which just checks for correctness rather than performance).

markccchiang:
No problem. I will update and merge my algorithm test once this PR is merged. My algorithm tester includes the time spending and the accuracy of percentile to pixel value under different algorithm settings. If there are other items you want to test, please let me know.

confluence:
A note about the cache: this PR changes the format of the keys (to incorporate one more field -- the base per-frame transformation used to calculate the quantile). This does not require a change in the database schema, but all existing database entries will never be used again, so it's a good idea to delete old databases and create new ones once this is merged.

grimmerk:
@confluence, I'm going to merge this pull request. Before that, I'd like to confirm some things with you. 

1. > Some unit conversions are unsupported, e.g. to/from Jy/pixel

Unsupport means it is possible but not support now, or this conversion is not possible?

2. >Frame-dependent beam information is not supported

I guess it is possible but very complicated, so not support now?

3. 
> Unit conversion is not supported in the pixel pipeline, which affects frame-dependent conversions;
> so changing units has no effect on how the image is rendered.

Does it mean that all frame-dependent conversions will still show incorrect frame image in imageViewer? If so, should we implement in the future or it does not matter for a scientifist? (I don't know) 

4. > can handle frame-dependent unit conversions (currently this is only implemented for conversions to and from Kelvin

Since frame-dependent unit conversions may have different percentile/quantile position, so does it mean that `to and from Kelvin` case will *loop all the data*  after convert all the data, then re-calculate the percentile/quantile again? If so, @markccchiang said that he did not see any code in this PR may call his percentile approximation after conversion. 

4.2 If this PR convert all the data for Kelvin, does it use the same casa API to loop all the data? Since that API will need to load the whole file to the memory, so just to confirm this point and we will find a day to overcome the memory bottleneck issue. 

4.3 Is it fast to convert all the data? 

4.4. Besides 4.2 question, will it consume a lot additional memory when converting all data? Since you mention `recursive call`, so just to confirm this point. 

5. From this table, https://docs.google.com/spreadsheets/d/1s9cY8iM4MHT3jI2VxIOVMrzTAF9MKNoN9BekOY-Whyo/edit#gid=0, 
Jy/pixel to Kelvin is possible? or it is an incorrect cell. @kswang1029 , any comment? 

6. From this doc, https://docs.google.com/document/d/1VZ64s2V47mJzXX4YBBTDN6jORu9XQKc6WFy3QeCdviU/edit, *Jy/beam -> Kelvin* is not possible 
> 2. frequency dependent, usually frequency changes monotonically in the cube (usual radio/mm interferometric data)

I can not understand, the above sentence seems to be a positive thing to support conversion.

grimmerk:
@kswang1029 , 

Since @confluence already writes some unit tests, do we need to verify by using other tools, such as using casa viewer to verify it? If so, either you (if you have time) or @markccchiang can help to verify it (you can tell him where the conversion function in casa viewer is).

kswang1029:
From this table, https://docs.google.com/spreadsheets/d/1s9cY8iM4MHT3jI2VxIOVMrzTAF9MKNoN9BekOY-Whyo/edit#gid=0,
Jy/pixel to Kelvin is possible? or it is an incorrect cell. @kswang1029 , any comment?
===> my mistake, Jy/pixel to Kelvin is not possible. cell corrected. 

From this doc, https://docs.google.com/document/d/1VZ64s2V47mJzXX4YBBTDN6jORu9XQKc6WFy3QeCdviU/edit, Jy/beam -> Kelvin is not possible
===> Jy/beam -> Kelvin is possible but we cannot convert the min/max of Jy/beam directly to the min/max of Kelvin since this conversion is frame-dependent (beam size and/or frequency).
The document just separates the two cases: 1) for those units that can be converted directly (e.g. Jy/arcsec^2 to/from MJy/sr) and 2) for those units that cannot be converted directly (e.g. Jy/beam to/from Kelvin).

kswang1029:
@kswang1029 ,

Since @confluence already writes some unit tests, do we need to verify by using other tools, such as using casa viewer to verify it? If so, either you (if you have time) or @markccchiang can help to verify it (you can tell him where the conversion function in casa viewer is).
===> I can verify it. casaViewer does not have unit conversion for images. In casaViewer, unit conversion only works for profiler (for the y-axis).

confluence:
Hi, Grimmer

1. As far as I know, conversions to/from Jy/pixel are possible; they were just never implemented in the old code. If we can find a good source for the formula we can easily add this (and other conversions) to the conversion plugin. I consider this to be a new feature, not an outstanding issue, since this wasn't implemented before the refactoring.

2. Again, this was never supported in the old code. If there is a need for it, it should be possible to add it by creating more converter subclasses -- we may just need additional error checking for the required information in the file, and we'll need to decide what to do if a conversion requires multiple frame-dependent transformations -- e.g. determine if there is a performance difference between multiple function calls and one function call, and therefore whether we should chain the functions somehow or create a single function per combination. Again, this would be a new feature.

3. Currently no matter what units are selected, the image will reflect the clipping values as calculated using the raw image units, because there is no support in the pixel pipeline for performing the value comparison in different units. This makes no difference with frame-independent conversions, but it does make a difference with frame-dependent conversions, which change the image statistics and should result in a different rendered image.<br/><br/>I have no idea whether the difference is big enough to matter. The old way of converting percentile values (by storing some arbitrary frame value that would be used to calculate the frequency in Hz) was also incorrect, and that didn't appear to matter -- so I don't think that this issue is serious enough to hold up the other refactoring. I do think that it would be a good idea to evaluate how difficult it would be to fix this.<br/><br/>Conceptually the algorithm would be analogous to the algorithm for calculating percentiles from pixel values. The biggest problem is likely to be the performance impact of performing an additional conversion calculation on every value being rendered. It should be possible to optimise this in the same way as the pixels -> percentiles algorithm is optimised -- the clipping values used for the comparison could be recalculated for each frame with the reverse transformation, and the unmodified image values would be compared to these values.

4. The percentile algorithms all take an optional converter parameter. If it exists and has a frame-dependent transformation, the algorithms loop over the data a frame at a time and convert the values. The converted values are then used however the unconverted values would be used. Constant multipliers are ignored; they are applied in the calling code after caching (so that we only cache each base frame-dependent transformation, not every possible conversion). I have added this modification to the approximate algorithms as well (e.g. see the `if (converter && converter->frameDependent)` block in `percentile2pixels_approximation`). No matter what converter is used, the data is only looped over once. If a value calculated with that base transformation is already cached, it does not need to be recalculated.

    4.2. Yes, we use the same API to loop over the data. I create slices for individual frames using the slice API defined in CARTA, and use them to iterate over one frame at a time.

    4.3. I haven't done an extensive performance test, so I'm not sure exactly how much this slows down the calculation. It did not appear to be noticeable when I tested it, but there may be a larger impact with larger files. We currently either multiply or divide by the square of the hertz value -- we may be able to speed up the calculations by precalculating the square of the value instead of the value unsquared; I'm not sure if the compiler already caches that in some way.

    4.4. No, this doesn't take up any extra memory. We convert one value at a time as we iterate over the data, and then either store it (in the exact algorithm) or use it in some other way (the approximate algorithms). We don't store an extra copy of the entire image. The recursive call I mentioned is to calculate the min and max which are needed for the approximate algorithm if they are not found in the cache. There is no change in performance from the previous non-recursive code -- I just eliminated an additional function in `DataSource` for *only* calculating the min and max, which unnecessarily duplicated what the main intensity function could already do.

5. I believe this was a mistake. I think the conversion is *possible*, but it was never implemented in the code.

6. Yes, this is possible. This is one of the frame-dependent conversions (all conversions to and from Kelvin are frame-dependent). You can look at the intensity conversion plugin to see what conversions are currently implemented (and how they are done).

kswang1029:
just a note: 
frame-dependent conversion is needed, especially for ALMA data which happen to have different beam sizes along the frequency axis. For that kind of image cube, the header should have a beam "table", not just bmaj, bmin, and bpa keywords (for uniform beam size image cube.)

When selecting an image file with the file browser of casaViewer, beam information will be shown to the right of file list. If the image file has an uniform beam, bmaj, bmin, and bpa are reported. If the image file has non-uniform beam, only medium values are reported and it takes significant amount of time before the medium values show up. I guess casacore can handle images with a beam "table".

confluence:
@kswang1029 We should be able to handle something like that by extending the converter code, in a similar way to how the frequency-based frame-dependent conversions are currently done. I guess it would be cleanest and simplest to create a separate converter for each direct conversion (i.e. one function per each possible combination of frame-dependent conversions).

kswang1029:
Jy/pixel to Kelvin is possible?
===> to be specific, Jy/pixel to Kelvin is possible only if we assume the beam size and shape is identical to the pixel. However, this is not an usual case.
Kelvin to Jy/pixel is possible only if we assume the beam size and shape are identical to the pixel size and shape (not necessary square pixel). Again, this is not an usual case. 
So I would say we should not allow Jy/pixel <-> Kelvin conversion.

just a note: Jy/pixel image can be made by deconvolving a Jy/beam image with its gaussian beam.

grimmerk:
@confluence, @markccchiang , @kswang1029 

I will merge this PR tonight. If you have any concern, please tell me.

kswang1029:
@grimmer0125 thanks and let me know which travis build is going to be tested. I will try to make some test images for this test.

PR #201, opened 2017-09-04
Cjhsu/histogram

appolloford:
This PR mainly contains three parts:
1) The histogram do not work for some images. This is caused by returning double type as int type.
2) Fix the wrong scale of histogram and use the default scale of qwt to draw.
3) The number of bins were usually reset by CARTA. Now it can work under the condition that the bin width is precise to the significant digit.
// Two commits were picked from cjhsu/profiler.


kswang1029:
Thanks for fixing some issues of histogram. Now histogram is shown for one of the test images which resulted in no histogram before. The bin number is correct and the displayed x-range is correct. 

There are still some synchronization issues in histogram. We can deal with those later. 

When animating a image cube with histogram derived per frame, there is a chance that the y-scale is corrupted due to very small values (perhaps related to logarithm). We can deal with this later too.

markccchiang (approved):
No problems find.

PR #210, opened 2017-09-19
Fix the bug from mark/addTimer for histogram

markccchiang:
Actually, this is the bug from my previous branch `mark/addTimer` which causes the fail in showing histogram plot. It is fixed now.


ajm-ska:
Confirmed to work on Linux.

ajm-ska:
It appears to work on Mac also.

PR #220, opened 2017-10-12
Cjhsu/refactor contour

appolloford:
This PR only do some refactor to remove the duplicate codes of contour


markccchiang (approved):
Good job! There should be no problem in this refactoring code. One thing that I found is the elapsed time for the refactoring `conrecFaster()` is a little longer by ~12% compared to the original code. It is happed for `Gaussian blur` or `Box blur` modes. But it may be not so obvious from user's experience if we consider the total time of plotting contours which includes the rendering.

grimmerk:
@appolloford how do you think?

markccchiang:
In the last commit 0cd99081456802c7f516baeb686227218ca1c6bb, @appolloford has improved the speed of plotting contour as well as the original version. So I think it is good to be merged.

appolloford:
If anyone needs, I can add more comments. Otherwise, I think this PR can be merged.

PR #221, opened 2017-10-16
Functional tester for approximate percentile algorithm

markccchiang:
This main program is used to test percentile algorithms, including finding Min/Max and approximate intensities with respect to percentiles. The usage is going to the build directory `$cartawork/CARTAvis/build/cpp/testPercentile` and executing the following command: `./testPercentile [the path and name of the image file you want to check] ...`. For example, type `./testPercentile /home/username/CARTA/Images/cartaImageLibrary/aH.fits`. 

The program will check Min/Max and approximate intensities with respect to percentiles, compare them with the precise value. The finding result of Min/Max intensity with different algorithms should be the same. On the other hand, if the absolute value of error for approximate intensity is smaller than the expected value, i.e., `1/[total number of bins]`, it will return `PASS`, otherwise, return `FAIL`. 

This tester checks the approximate algorithm for percentiles to intensities with the total bin number `1,000`, `10,000`, `100,000`, and `1,000,000`. For each case of the total bin number, it checks percentiles `0.0005`, `0.0025`, `0.005`, `0.01`, `0.015`, `0.02`, `0.025`, `0.0375`, `0.05`, `0.95`, `0.9625`, `0.975`, `0.98`, `0.985`, `0.99`, `0.995`, `0.9975`, and `0.9995`. The reason for returning `FAIL` is because the distribution of the pixel bin may be not continuous everywhere. It usually happens for the percentile very close to `0` or `1`.


grimmerk:
@confluence , 

@markccchiang told me `this tester` is not written by Qt's unit test framework. So there are two things,  
1. Is the testing code without Qt's unit test enough to cover the algorithms? 
2. Is Learning Qt's unit test framework valuable or any extra benefit? 
How do you think? I prefer to write Qt unit test's version but I am not familiar with it.

confluence:
Hello,

There is nothing particularly special about the Qt test framework; I used it for the intensity plugin test because it was convenient, but I used Catch for my own percentile algorithm test (in my branch, under Tests). I'm on leave today, but I'm back tomorrow, and will start working on merging my outstanding changes.

grimmerk:
@markccchiang ,  Please change to use `./carta/scriptedClient/tests/data/aH.fits` instead.

markccchiang:
Done.

PR #222, opened 2017-10-25
Hard code colormaps in C++ syntax

markccchiang:
The CARTA colormaps from Matplotlib Python, i.e. `ColormapsPy` plugin, are "Spectral", "coolwarm", "Set1", "gnuplot2", "jet", "Reds", "gist_stern", "Blues", "terrain", "gnuplot", "hot", "Greens" and "nipy_spectral". In this PR, I just hard code these colormaps as C++ syntax in `Colormaps1` plugin and rename them as  "Spectral-test", "coolwarm-test", "Set1-test", "gnuplot2-test", "jet-test", "Reds-test", "gist_stern-test", "Blues-test", "terrain-test", "gnuplot-test", "hot-test", "Greens-test" and "nipy_spectral-test". Users can test the difference of the same colormap definition from different sources. The definition of colormaps from Matplotlib can be found in the Python library `/usr/local/lib/python2.7/site-packages/matplotlib/_cm.py` if users have installed Matplotlib on their computers.

The advantage of hard-coding these colormaps in the CARTA is that we don't need to rely on the Matplotlib Python package. Also, it can increase the speed of rendering the image by ~5 times compared to the original codes. It is especially sensible while users playing the image animation at the first cycle run. If there are no problems found from tests, I would expect to remove the `ColormapsPy` plugin and rename the colormap "XXX-test" back to "XXX".


markccchiang:
Now I remove python plugins, "python273" and "ColormapsPy", and add more "cubeHelix" colormaps. There is an error in the original "cubeHelix" function code, I have correct it. The new name of "cubeHelix" has four parameters in it, which are (`start`, `rots`, `hue`, `gamma`). Users can easily tell values of these parameter settings from the name of "cubeHelix" colormaps.

grimmerk:
@markccchiang , after thinking, it is better to keep an Python plugin example for demonstration, if someone wants to write any Python plugin.

markccchiang:
I have restored the "python273" plugin.

PR #223, opened 2017-10-26
Remove the html location bar on the top of UI panel

markccchiang:
This is the side effect of my previous PR branch `mark/remove-qtDecorations-opt`. It happens for the Mac but not on the Linux system.


grimmerk:
mark/remove-qtDecorations-opt: #219.

Issue #142, opened 2017-05-05
statistics of a region


kswang1029:
The statistics of a region always adopts the pixels of the first channel (channel index=0). If the current channel index is 3 and a region is defined, we should see statistics of channel index 3. 

There should also an option that when a region defined, there should be a capability to specify this region is applied to current channel or the whole channels (or a range of channels). Therefore the statistics can be calculated based on the "region".

<img width="1279" alt="screen shot 2017-05-05 at 2 09 04 pm" src="https://cloud.githubusercontent.com/assets/20819712/25735495/62db2796-319e-11e7-8e7d-33443835f7c0.png">

PR #226, opened 2017-11-04
Correct statistical information for different channels and regions

markccchiang:
This PR solves the first item of issue #142. The statistics information now can be updated for the selecting region including the current spectral or stoke channel. The `Rectangle` region performs well. However, the other region types, `Polygon`, `Ellipse` and `Point` need further checks.

### DIFF
--- a/carta/cpp/core/Data/Image/LayerGroup.cpp
+++ b/carta/cpp/core/Data/Image/LayerGroup.cpp
@@ -242,8 +242,12 @@ void LayerGroup::_displayAxesChanged(std::vector<AxisInfo::KnownType> displayAxi
     }
 }
 
-Carta::Lib::AxisInfo::KnownType LayerGroup::_getAxisType( int /*index*/ ) const {
+Carta::Lib::AxisInfo::KnownType LayerGroup::_getAxisType( int index ) const {
     AxisInfo::KnownType axisType = AxisInfo::KnownType::OTHER;
+    int dataIndex = _getIndexCurrent();
+    if ( dataIndex >= 0 ){
+        axisType = m_children[dataIndex]->_getAxisType( index );
+    }
     return axisType;
 }
 
--- a/carta/cpp/core/Data/Image/LayerGroup.cpp
+++ b/carta/cpp/core/Data/Image/LayerGroup.cpp
@@ -244,10 +244,10 @@ void LayerGroup::_displayAxesChanged(std::vector<AxisInfo::KnownType> displayAxi
 
 Carta::Lib::AxisInfo::KnownType LayerGroup::_getAxisType( int /*index*/ ) const {
     AxisInfo::KnownType axisType = AxisInfo::KnownType::OTHER;
-    //int dataIndex = _getIndexCurrent();
-    //if ( dataIndex >= 0 ){
-    //    axisType = m_children[dataIndex]->_getAxisType( index );
-    //}
+    int dataIndex = _getIndexCurrent();
+    if ( dataIndex >= 0 ){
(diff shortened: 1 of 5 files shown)


markccchiang:
As my understanding, when we get `Rectangle`, `Polygon`, `Ellipse` or `Point` region, `Rectangle` and `Point` objects can select the range of spectral or stoke channels (frames). By default, we select the current view of the channel. They are defined in "RegionRecordFactory.cpp". On the other hand, `Polygon` and `Ellipse` objects cannot select the range of spectral or stoke channels (by default they select all channels). So we need the second stage of the "box region" process, which is used to select the range of spectral or stoke channels.

markccchiang:
TODO: create a UI panel that allows users to select a specific range of spectral channels for statistics. Current statistics information is only on the current channel from image viewer.

carta/cpp/core/Data/Image/LayerGroup.cpp
```diff
@@ -242,8 +242,12 @@ void LayerGroup::_displayAxesChanged(std::vector<AxisInfo::KnownType> displayAxi
     }
 }
 
-Carta::Lib::AxisInfo::KnownType LayerGroup::_getAxisType( int /*index*/ ) const {
+Carta::Lib::AxisInfo::KnownType LayerGroup::_getAxisType( int index ) const {
     AxisInfo::KnownType axisType = AxisInfo::KnownType::OTHER;
+    int dataIndex = _getIndexCurrent();
+    if ( dataIndex >= 0 ){
+        axisType = m_children[dataIndex]->_getAxisType( index );
```
grimmerk: Why does this affect the whole behavior?

grimmerk (commented):
Some question.

carta/cpp/core/Data/Image/LayerGroup.cpp
grimmerk: Does _getAxisType have some kind of setter behavior inside?

grimmerk: Also, the original implementation of this function is empty? Is it wrong?

markccchiang: The original implementation of this function is empty and it would cause problems at least for statistical function. We have to get the current image data `dataIndex`, i.e. get the current layer, from `LayerGroup.cpp` and then get the axis type via `_getAxisType()` of the current layer from `LayerData.cpp`. It is the same way as we applying `_getAxisXType()` and `_getAxisYType()` functions.

markccchiang: The original implementation of this function is empty and it would cause problems at least for statistical function. We have to get the current image data `dataIndex`, i.e. get the current layer, from `LayerGroup.cpp` and then get the axis type via `_getAxisType()` of the current layer from `LayerData.cpp`. It is the same way as we applying `_getAxisXType()` and `_getAxisYType()` functions.

carta/cpp/core/Data/Image/LayerGroup.cpp
```diff
@@ -244,6 +244,10 @@ void LayerGroup::_displayAxesChanged(std::vector<AxisInfo::KnownType> displayAxi
 
 Carta::Lib::AxisInfo::KnownType LayerGroup::_getAxisType( int /*index*/ ) const {
     AxisInfo::KnownType axisType = AxisInfo::KnownType::OTHER;
+    //int dataIndex = _getIndexCurrent();
+    //if ( dataIndex >= 0 ){
+    //    axisType = m_children[dataIndex]->_getAxisType( index );
+    //}
```
markccchiang: Could be uncommented.

PR #229, opened 2017-11-09
memory issue (not work in permutation mode which will be improved in another PR)

appolloford:
The cause of current limit of memory may come from the improper setting of RO_LatticeIterator. I changed the cursorshape so that it seems that CARTA won't load whole file when running forEach function.

### DIFF
--- a/carta/cpp/plugins/CasaImageLoader/CCRawView.h
+++ b/carta/cpp/plugins/CasaImageLoader/CCRawView.h
@@ -8,6 +8,7 @@
 #include <casacore/lattices/Lattices/LatticeStepper.h>
 #include <casacore/lattices/Lattices/LatticeIterator.h>
 #include <casacore/casa/Arrays/IPosition.h>
+#include <algorithm>
 
 template < typename PType >
 class CCImage;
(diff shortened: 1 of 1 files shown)


carta/cpp/plugins/CasaImageLoader/CCRawView.h
```diff
@@ -219,12 +220,20 @@ CCRawView < PType >::forEach(
...
-    /// of size 2x1(x1x1x1....x1) regardless of the image size...
-    auto cursorShape = imageShape;
+    // set the cursor shape to load partial file at each step
+    casacore::IPosition cursorShape( imgDims );
+    auto shapeVec = imageShape.asVector();
+    for(auto i=0; i<imgDims; i++){
+        // Restrict the maximum size in one dim to 16
+        // (The number is given arbitarily. Maybe it can be changed in the future)
+        cursorShape(i) = std::min( shapeVec(i), 16);
+        if ( i == 0 || i == 1 ){
```
grimmerk, line 230: 1. Does this PR work in the permutation case? 
2. In permutation mode, does the copied permutated image still occupy the same size?

grimmerk (commented):
Some questions.

PR #230, opened 2017-11-13
Change the tool bar name "Data" to "File"

markccchiang:
It may be more intuitive for users to start working on the application program by chipping "File" button.


grimmerk:
Since this point is already discussed, so close it.

PR #241, opened 2017-12-12
Update colored histogram with different percentile or colormap settings

markccchiang:
Let colored histogram can be updated with different percentile or colormap settings, as shown below:

### DIFF
--- a/carta/cpp/core/Data/Image/Controller.cpp
+++ b/carta/cpp/core/Data/Image/Controller.cpp
@@ -88,6 +88,7 @@ Controller::Controller( const QString& path, const QString& id ) :
 	connect( m_stack.get(), SIGNAL(contourSetRemoved(const QString&)),
 			this, SLOT(_contourSetRemoved(const QString&)));
 	connect( m_stack.get(), SIGNAL(colorStateChanged()), this, SLOT( _loadViewQueued() ));
+    connect( m_stack.get(), SIGNAL(colorStateChanged()), this, SLOT(_emitColorChanged()));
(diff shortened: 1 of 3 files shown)


carta/cpp/core/Data/Image/Controller.cpp
```diff
@@ -88,6 +88,7 @@ Controller::Controller( const QString& path, const QString& id ) :
 	connect( m_stack.get(), SIGNAL(contourSetRemoved(const QString&)),
 			this, SLOT(_contourSetRemoved(const QString&)));
 	connect( m_stack.get(), SIGNAL(colorStateChanged()), this, SLOT( _loadViewQueued() ));
+    connect( m_stack.get(), SIGNAL(colorStateChanged()), this, SLOT(_emitColorChanged()));
```
markccchiang: Can be simplified like the following code line.

carta/cpp/core/Data/Image/Controller.cpp
```diff
@@ -87,7 +87,7 @@ Controller::Controller( const QString& path, const QString& id ) :
 			this, SLOT(_contourSetAdded(Layer*, const QString&)));
 	connect( m_stack.get(), SIGNAL(contourSetRemoved(const QString&)),
 			this, SLOT(_contourSetRemoved(const QString&)));
-	connect( m_stack.get(), SIGNAL(colorStateChanged()), this, SLOT( _loadViewQueued() ));
+    connect( m_stack.get(), SIGNAL(colorStateChanged()), this, SLOT(_emitColorChanged()));
```
markccchiang, line 90: If we change the colormap setting on the UI. The colored histogram will also update to the new colormap.

PR #242, opened 2017-12-15
Label the colormap bar with intensities

markccchiang:
Label the colormap bar with intensities as shown in the figure below. The intensity label can be changed correctly with different percentile or unit conversion settings. The similar way can be applied to the new CARTA framework.


carta/cpp/core/Data/Colormap/Colormap.cpp
```diff
@@ -163,10 +166,35 @@ void Colormap::_calculateColorStops(){
...
+                    float val_for_label = intensityMin_for_label + j*delta_for_label;
+
+                    // fill intensities in a string list
+                    QString sci_val_for_label = QString::number(val_for_label, 'E', 3);
+                    if (j < number_of_section) {
+                        sci_val_for_label = sci_val_for_label + ",";
+                    }
+                    buff_val_for_label.append(sci_val_for_label);
+                }
+                m_state.setValue<QString>( COLOR_GRADES, buff_val_for_label.join("") );
```
markccchiang: Maybe not a correct way to transform the unit.

PR #245, opened 2018-01-18
Cjhsu/new arch grid control

appolloford:
This PR discard the original design of gridcontrol, and use the controller to receive the commands from the front-end, and then call stack to deal with the requests. The purpose is to solve the previous sync problem. We don't need to sync the datagrid belonging to the gridcontrol and another one belonging to  the layerdata. The front-end will directly get the state of the datagrid in the layerdate to set the props of the UI elements.
Details:
1. There are some commands still not be transferred (related to the color setting) so the definition of gridcontrol is left as a reference.
2. The transferred commands can be found in the commit description. The left commands also can be found in the comment of controller.
3. The related part in the python interface is disabled/commented.
4. Most of the 'set' function in the datagrid are replaces by the '_setState()', the function will directly set the state value of the datagrid and return the updated state string. Something noticeable is that the setState won't check the value. The check should be done in the front-end. Besides, the '_setState' function doesn't handle exception, which should be added in the future. The original 'set' functions are also left as references.
5. In the PR, the setDataGridState method in LayerGroup will change the datagrid states of all the LayerData members and return the datagrid state of the current layerdata. The behavior should be discussed further.

### DIFF
--- a/carta/cpp/core/Data/Image/Controller.cpp
+++ b/carta/cpp/core/Data/Image/Controller.cpp
@@ -1062,13 +1062,8 @@ void Controller::_initializeCallbacks(){
     //
     addCommandCallback( "setCoordinateSystem", [=] (const QString & /*cmd*/,
             const QString & params, const QString & /*sessionId*/) -> QString {
-        // std::set<QString> keys = {DataGrid::COORD_SYSTEM};
-        // std::map<QString,QString> dataValues = Carta::State::UtilState::parseParamMap( params, keys );
-        // QString showAxisStr = dataValues[DataGrid::SHOW_AXIS];
-        // QString result = setCoordinateSystem( dataValues[DataGrid::COORD_SYSTEM] );
-        // Util::commandPostProcess( result );
-        QString stateName = DataGrid::COORD_SYSTEM;
-        QString result = m_stack->_setDataGridState( stateName, params );
+
+        QString result = m_stack->_setCoordinateSystem( params );
         return result;
     });
     //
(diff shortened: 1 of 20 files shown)


carta/cpp/core/Data/Image/Controller.cpp
```diff
@@ -1451,17 +1768,17 @@ void Controller::_updateCursorText(bool notifyClients ){
 
 
 void Controller::_updateDisplayAxes(){
```
grimmerk: comment this part.

carta/cpp/core/Data/Image/DataSource.cpp
```diff
@@ -136,11 +136,23 @@ std::vector<AxisInfo::KnownType> DataSource::_getAxisTypes() const {
...
+
+    if ( cs == Carta::Lib::KnownSkyCS::Unknown ) {
+        qWarning() << "Unknown celestial coordinate system.";
+    }
+    else if ( cs == Carta::Lib::KnownSkyCS::Error ){
+        qCritical() << "Error celestial coordinate system.";
+        CARTA_ASSERT("There are something wrong when trying get axisinfo.");
+    }
+    else if ( cs != Carta::Lib::KnownSkyCS::Default ) {
+        cf->setSkyCS( cs );
```
grimmerk: people may feel confused about _getAxisInfos  & cf->setSkyCS( cs );? One is getter, the other one is setter fun.

carta/cpp/core/Data/Image/Grid/GridControls.cpp
```diff
@@ -53,6 +53,14 @@ void GridControls::_initializeDefaultState(){
 
 void GridControls::_initializeCallbacks(){
 
+    addCommandCallback( "getDataGridState", [=] (const QString & /*cmd*/,
```
grimmerk: can be removed. useless?

PR #246, opened 2018-01-23
Move approximate quantile algorithm to plugin; add another approximate algorithm

confluence:
Hi, everyone

This pull request moves the histogram-based percentile approximation into a plugin, and adds another approximation as a different plugin. If multiple plugins are enabled, currently the one which reports the smallest error is selected. The second plugin is experimental, and disabled by default in the config. If no approximate plugin can be loaded, the exact algorithm is used. Disabling both plugins disables approximate calculations entirely.

Here is a summary of the changes:

* The configuration has changed: the global parameter used to configure the histogram approximation is now a plugin parameter. The second plugin also requires parameters.
* Interfaces for pixel<->percentile calculations were created in CartaLib.
* The exact algorithms in the core and the approximate algorithms in the plugins implement these interfaces.
* Unit tests for the approximate algorithms were moved to their respective plugins.
* `testPercentile` was refactored to load the plugins. It runs the tests on both plugins, but as I haven't yet established the expected error of the experimental algorithm, the test only prints the errors for various parameters instead of doing a comparison. The test does not currently override the config to enable all plugins, so disabling the plugin disables these additional tests.

This PR touches a lot of things -- I think I've correctly synced to the latest development branch, but please let me know if further corrections are required.


markccchiang:
Thanks for your remarkable works! I am reading your codes and we will merge it if there is no problem.

markccchiang:
There is no problem for me for this PR. The error estimation for `Manku99` algorithm is not available yet, so if I turn on this experimental plugin it will lead to the crash of the CARTA (while opening an image file). You will complete this plugin function, right? Also, you plan to add an option that allows CARTA can choose `histogram percentile` or `Manku99` automatically?

confluence:
Nothing should crash. The plugin works; it's just not clear what the error is. A dummy error value is used. There is already effectively an option for selecting which plugin is used -- the disused one should be disabled (in the same why that was done for the cache plugin). The intended mode of operation is that the plugin with the lowest error is used -- this may vary depending on the image, but at the moment each plugin returns a constant error. This can be refined later to almost any criterion.

carta/cpp/core/Data/Image/DataSource.cpp
```diff
@@ -426,158 +428,138 @@ std::shared_ptr<Carta::Core::ImageRenderService::Service> DataSource::_getRender
...
-
-        // define the priority number (error order) for choosing (key, value) from SQLite3
-        double error;
+        // if the algorithm is approximate, add extra percentiles from clips, but only if they are not cached
         
+        if (calculator->isApproximate) {
+            std::shared_ptr<Carta::Data::Clips> m_clips;
+            std::vector<double> percentilesFromClips = m_clips->getAllClips2percentiles();
+            
+            for (auto& p : percentilesFromClips) {
```
markccchiang, line 540: The crash problem is due to the total number of percentiles to calculate are not matched with the expected number in `ManKu99` function, i.e., in `PercentileManku99.h` line 549. To solve this, I would change this "for loop": `for(auto& p : percentilesFromClips) { ... }` as the following:
```
            for (auto& p : percentilesFromClips) {
                // TODO check exactly why this is necessary
                // check if extra percentiles are close to any existing percentiles, cached or uncached
                // C.C.Chiang: For approximation method, we can calculate all clipping values (all buttons) which are
                // listed on the UI panel at the same time. The following code block is used to fill in all clipping
                // values (as a set) we want to calculate.
                bool isDuplicate = false;
                for (size_t i = 0; i < percentiles.size(); i++) {
                    if (fabs(percentiles[i] - p) < 1e-6) {
                        isDuplicate = true;
                        // Either this was found in the cache already
                        // Or it's already in the list of percentiles to be calculated
                        // Either way, ignore it
                        break;
                    }
                }
                if (!isDuplicate) {
                    // This is a different percentile
                    // Look in the cache first
                    std::shared_ptr<Carta::Lib::IntensityValue> cachedValue = _readIntensityCache(frameLow, frameHigh, p, stokeFrame, transformationLabel);
                    // Add it to the list if it's not in the cache
                    if (!cachedValue) {
                        percentilesToCalculate.push_back(p);
                    }
                }
            }
```

confluence: Oh, I see what's happening -- I forgot that this algorithm eliminates duplicates. I will fix it.

confluence:
I believe that simply removing the check is the correct thing to do here.

confluence:
I had to figure out why debugging output was being suppressed on my machine (it's [an Ubuntu bug](https://bugs.launchpad.net/ubuntu/+source/qtbase-opensource-src/+bug/1731646)), but I have now examined the output with the Manku99 plugin enabled and it appears to be working correctly (except that the error value is a dummy value).

markccchiang:
Thanks! It should be no problem now. We will review the code and merge it. Since Grimmer just left last week. I think we will keep the current CARTA framework for a while beyond our original expectation.. before it can be completely migrated to the new architecture.

carta/cpp/plugins/PercentileManku99/PercentileManku99.h
```diff
@@ -0,0 +1,557 @@
...
+
+private:
+    size_t numBuffers;
+    size_t bufferCapacity;
+    size_t sampleAfter;
+};
+
+// TODO: error is completely wrong; work out what it actually is
+template <typename Scalar>
+PercentileManku99<Scalar>::PercentileManku99(const size_t numBuffers, const size_t bufferCapacity, const size_t sampleAfter) :Carta::Lib:: IPercentilesToPixels<Scalar>(0.5, "Manku99 approximation", true), numBuffers(numBuffers), bufferCapacity(bufferCapacity), sampleAfter(sampleAfter) {
```
markccchiang, line 490: Hi, I have one more question. Currently, the error of Manku99 algorithm is set to be a fixed value 0.5 (50%). In theory, the error of this algorithm is less than `numBuffers/bufferCapacity`. But in practice, we are not sure is it correct or not. Is my understanding correct?

confluence:
The algorithm is non-deterministic, and I haven't worked out what a sensible error value should be yet. 0.5 is a dummy value.

PR #248, opened 2018-01-31
[newArch] add clipping state functions for the Meteor

markccchiang:
The new functions can transfer the current clipping state to the Meteor client site and save it in the MongoDB. This PR is corresponding to the PR in `CARTAvis/newCARTAMeteorApp #16`.

### DIFF
--- a/carta/cpp/core/Data/Image/Controller.cpp
+++ b/carta/cpp/core/Data/Image/Controller.cpp
@@ -620,7 +620,23 @@ void Controller::_initializeCallbacks(){
         return result;
     });
 
+    // get the Clipping state
+    addCommandCallback( "getClipState", [=] (const QString & /*cmd*/,
+                const QString & /*arg*/, const QString & /*sessionId*/) -> QString {
+        QString result;
+        result = getClipState();
+        Util::commandPostProcess( result );
+        return result;
+    });
 
+    // get the Auto Clipping state
+    addCommandCallback( "getAutoClipState", [=] (const QString & /*cmd*/,
+                const QString & /*arg*/, const QString & /*sessionId*/) -> QString {
+        QString result;
+        result = getAutoClipState();
+        Util::commandPostProcess( result );
+        return result;
+    });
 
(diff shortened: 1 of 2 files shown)


carta/cpp/core/Data/Image/Controller.cpp
```diff
@@ -1200,6 +1215,28 @@ void Controller::recallClipValue() {
     double minPercent = m_state.getValue<double>(CLIP_VALUE_MIN);
     double maxPercent = m_state.getValue<double>(CLIP_VALUE_MAX);
     emit clipsChanged( minPercent, maxPercent, autoClip );
```
grimmerk, line 1217: @appolloford it may result in some unnecessary slot/side effect?

carta/cpp/core/Data/Image/Controller.cpp
```diff
@@ -1155,10 +1171,9 @@ void Controller::setAutoClip( bool autoClip ){
     if ( autoClip != oldAutoClip ) {
         m_state.setValue<bool>( AUTO_CLIP, autoClip );
         m_state.flushState();
+        // refresh the image viewer
+        recallClipValue();
```
grimmerk, line 1175: what is the effect of this?

grimmerk (commented):
Please check some points.

carta/cpp/core/Data/Image/Controller.cpp
markccchiang: The `emit clipsChanged(...)` triggers the recalculation of the percentile w or w/o per frame. If I remove the `recallClipValue()` the image cannot update with different auto clip settings.

carta/cpp/core/Data/Image/Controller.cpp
markccchiang: `recallClipValue()` is used to update the image with different auto clip settings.

markccchiang: Check duplicate behavior of `emit clipsChanged(...)`.

